PR #79, opened 2019-08-27
Added data filtering to TablePlotter

HansKallekleiv:
Added filtering option to the TablePlotter.
The filter is activated by adding the argument `filter_cols` to the config files, and specifying a list of dataframe columns that are filterable.

For numerical columns a RangeSlider is rendered. For other datatypes a dropdown is rendered.

### DIFF
--- a/examples/basic_example.yaml
+++ b/examples/basic_example.yaml
@@ -40,7 +40,11 @@ pages:
    content:
     - container: TablePlotter
       csv_file: ./example_data.csv
-      filter_cols: ['Well', 'Segment', 'Average permeability (D)']
+      filter_cols:
+          - Well
+          - Segment
+          - Average permeability (D)
+          - ss
       contact_person:
         name: Ola Nordmann
         phone: +47 12345678
--- a/examples/basic_example.yaml
+++ b/examples/basic_example.yaml
@@ -44,7 +44,6 @@ pages:
           - Well
           - Segment
           - Average permeability (D)
-          - ss
       contact_person:
         name: Ola Nordmann
         phone: +47 12345678
--- a/examples/basic_example.yaml
+++ b/examples/basic_example.yaml
@@ -40,6 +40,7 @@ pages:
    content:
     - container: TablePlotter
       csv_file: ./example_data.csv
+      filter_cols: ['Well', 'Segment', 'Average permeability (D)']
       contact_person:
         name: Ola Nordmann
         phone: +47 12345678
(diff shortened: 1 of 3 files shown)


webviz_config/containers/_table_plotter.py
```diff
@@ -188,12 +248,22 @@ def style_options_div_hidden(self):
...
+                'justify-content': 'space-between',
+                'grid-template-columns': '1fr 5fr 1fr'
+            }
+        else:
+            return {
+                'display': 'grid',
+                'align-content': 'space-around',
+                'justify-content': 'space-between',
+                'grid-template-columns': '1fr 5fr'
+            }
```
anders-kiaer: These two are quite equal. Change to

1) Define dictionary outside if/else, except grid-template-columns.
2) Define grid-template-columns value using inline if-else

perhaps.

examples/basic_example.yaml
```diff
@@ -40,6 +40,7 @@ pages:
    content:
     - container: TablePlotter
       csv_file: ./example_data.csv
+      filter_cols: ['Well', 'Segment', 'Average permeability (D)']
```
anders-kiaer: Is this the best way, or should we demonstrate the other way of defining lists in yaml (i.e. one col per line). Verbosity vs «less-code-looking» config.

tests/test_table_plotter.py
```diff
@@ -43,6 +45,46 @@ def test_table_plotter(dash_duo):
...
+    assert plot_dd.text == 'scatter'
+
+    # Checking that only the relevant options are shown
+    for plot_option in page.plot_args.keys():
+        plot_option_dd = dash_duo.find_element(
+            f'#{page.plot_option_id}-div-{plot_option}')
+        if plot_option not in page.plots['scatter']:
+            assert plot_option_dd.get_attribute('style') == 'display: none;'
+        else:
+            assert plot_option_dd.get_attribute('style') == 'display: grid;'
```
anders-kiaer: Flip if/else and remove `not`?

webviz_config/containers/_table_plotter.py
```diff
@@ -132,6 +147,50 @@ def plot_args(self):
...
+                                id=self.filter_ids[col],
+                                options=[{'label': i, 'value': i}
+                                         for i in elements],
+                                     value=elements,
+                                     multi=True)
+                        ])
+                    ])
+                )
+        return dropdowns
+
```
anders-kiaer: This fails if filter column is constant? I.e. min_val = max_val.

HansKallekleiv: Fixed by removing constant columns from the filter.

PR #102, opened 2019-09-26
Add progress information when making portable webviz

### LINKED ISSUE #85: Improved status print when creating portable instance
When creating a portable instance, the different functions called can be time consuming and many.
The user can sometimes be unsure if it is still running and/or overall progress

E.g.:
 - Add a dynamic "terminal spinner" to show the script is still live.
 - Add overall progress (e.g. finisihed with x% of the functions to be called).

anders-kiaer:
Resolves #85.

### DIFF
--- a/webviz_config/webviz_store.py
+++ b/webviz_config/webviz_store.py
@@ -145,6 +145,12 @@ class WebvizStorage:
             )
 
     def build_store(self):
+
+        total_calls = sum(
+            len(calls) for calls in self.storage_function_argvalues.values()
+        )
+        counter = 0
+
         for func in self.storage_functions:
             for argtuples in self.storage_function_argvalues[func]:
                 kwargs = dict(argtuples)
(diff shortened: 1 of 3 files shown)


webviz_config/webviz_store.py
```diff
@@ -167,7 +173,10 @@ def build_store(self):
                 else:
                     raise ValueError(f"Unknown return type {type(output)}")
 
-                print(" \033[92m\033[1m[\u2713] Saved\033[0m")
+                counter += 1
```
HansKallekleiv: Not sure if it is better but could get counter by
`for counter, argtuples in enumerate(self.storage_function_argvalues[func]):`

anders-kiaer: That would have been better, but there is an outer loop as well which prevents tha (outer is on the different functions to call, inner on the different arg combinations to a given function).

PR #119, opened 2019-10-09
Disallow CSP style-inline

anders-kiaer:
This is a test to see if we can also disallow CSP style-inline. Most of the hashes are needed due to [dash-core-components](https://github.com/plotly/dash-core-components) adding inline-style tags dynamically (element style attributes are not a problem). I.e. the best solution would be that `dash-core-components` load a static CSS file + a static JS file instead.

**Edit:** This PR is on hold - trying to make a PR in the external `dash-core-components` repo. instead.


anders-kiaer:
This issue is now raised in `das-core-components` and will likely be solved there instead.

PR #153, opened 2019-11-19
Resolve path to lists of pathlib objects

### LINKED ISSUE #154: Allow lists of paths as config argument
The yaml config parser currently resolves arguments with a single path to an absolute path before sending it to the container.

In many cases it is useful to have lists of paths as container argument.
```python
from typing import List
from pathlib import Path
def __init__(some_arg: List[Path])
...
```

The yaml parser should handle these lists of paths.

HansKallekleiv:
Closes #154

### DIFF
--- a/webviz_config/_config_parser.py
+++ b/webviz_config/_config_parser.py
@@ -110,7 +110,7 @@ def _call_signature(
 
             if expected_type == pathlib.Path:
                 kwargs[arg] = (config_folder / pathlib.Path(kwargs[arg])).resolve()
-            if expected_type == typing.List[pathlib.Path]:
+            elif expected_type == typing.List[pathlib.Path]:
                 kwargs[arg] = [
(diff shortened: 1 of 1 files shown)


webviz_config/_config_parser.py
```diff
@@ -109,17 +110,25 @@ def _call_signature(
...
-                raise ParserError(
-                    f"{terminal_colors.RED}{terminal_colors.BOLD}"
-                    f"The value provided for argument `{arg}` "
-                    f"given to container `{container_name}` is "
-                    f"of type `{type(kwargs[arg]).__name__}`. "
-                    f"Expected type "
-                    f"`{argspec.annotations[arg].__name__}`."
-                    f"{terminal_colors.END}"
-                )
+            if expected_type == typing.List[pathlib.Path]:
```
anders-kiaer: `elif`?

PR #157, opened 2019-11-22
Resolve scratch_ensemble paths

HansKallekleiv:
Closes https://github.com/equinor/webviz-subsurface/issues/128


anders-kiaer:
Discussed with @HansKallekleiv - we will solve it using a different approach keeping FMU out of `webviz-config`.

PR #152, opened 2019-11-19
Add temporary dash table workaround

HansKallekleiv:
Relates to #149

PR #161, opened 2019-11-28
Use dash~=1.7. Remove hidden dash table

HansKallekleiv:
[Dash 1.7.0](https://github.com/plotly/dash/releases/tag/v1.7.0) includes the fix for `https://github.com/plotly/dash/pull/1027`.

It is no longer necessary to have a hidden dash_table in the initial rendered layout.

- Bump dash to 1.7
- Remove hidden table
- Also bump webviz-core-components

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ setup(
     },
     entry_points={"console_scripts": ["webviz=webviz_config.command_line:main"]},
     install_requires=[
-        "dash~=1.1",
+        "dash>=1.7",
         "bleach~=3.1",
         "cryptography~=2.4",
         "flask-caching~=1.4",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -31,7 +31,7 @@
     },
     entry_points={"console_scripts": ["webviz=webviz_config.command_line:main"]},
     install_requires=[
-        "dash~=1.1",
+        "dash>=1.7",
```
anders-kiaer: Keep `~=` by using `dash~=1.7`?

webviz_config/templates/webviz_template.py.jinja2
```diff
@@ -107,12 +107,6 @@ else:
                 className="styledButton",
               {%- endif -%}
                 children=[
-                  {%- if loop.first -%}
-                    html.Div(
-                        id="dash-issue-#1010",
-                        style={"display":"none"}, 
-                        children=dash_table.DataTable()),
-                  {%- endif -%}
```
anders-kiaer: In #152 it was also [introduced](https://github.com/equinor/webviz-config/pull/152/files#diff-b5cac186109f24b6fad1fd3d8de0318fR17) a `import dash_table` which we perhaps also can remove now.

PR #162, opened 2019-11-28
Support third-party containers checking shared_settings

### LINKED ISSUE #160: General container settings
`webviz-config` has, in the configuration file, a dictionary called `container_settings` which are there for the user to set general configurations, shared between multiple containers.

The individual containers can ask for this dictionary, and take out the values they need (for a given predefined key). However - it would be useful if `webviz-config` had somewhat more knowledge about the structure of `container_settings`:

- If `webviz-config` knows the types of the different elements going in, it can do transformations (e.g. go from a string being a path relative to the configuration file, to an absolute `Path` instance).
- If the containers has some checks they want to perform on the dictionary `containers_settings`, they can do that early in the app startup process.
- `webviz-config` can potentially detect conflicts between different third-party containers requesting and (re)defining the same key in `container_settings`.

anders-kiaer:
Resolves #160.

Third-party containers can, in order to get rid of depreciation warning introduced here, do the following change:

Go from e.g.
```python
def __init__(self, app, container_settings, ...):
    something = container_settings["something"]
    ...
```
to
```python
def __init__(self, app, ...):
    something = app.webviz_settings["shared_settings"]["something"]
    ...
```

### DIFF
--- a/webviz_config/templates/copy_data_template.py.jinja2
+++ b/webviz_config/templates/copy_data_template.py.jinja2
@@ -5,10 +5,11 @@ import os.path as path
 from pathlib import Path, PosixPath
 
 import dash
+
+import webviz_config
 from webviz_config.webviz_store import WEBVIZ_STORAGE
 from webviz_config.webviz_assets import WEBVIZ_ASSETS
 from webviz_config.common_cache import CACHE
-
 
 
 {% for module in _imports %}
@@ -23,6 +24,7 @@ app = dash.Dash()
 app.config.suppress_callback_exceptions = True
 
 app.webviz_settings = {
+    "shared_settings": webviz_config.SHARED_SETTINGS_SUBSCRIPTIONS.transformed_settings({{ shared_settings }}, {{ config_folder }}),
     "plotly_layout": {{ plotly_layout }},
 }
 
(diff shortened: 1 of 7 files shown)


webviz_config/templates/copy_data_template.py.jinja2
```diff
@@ -23,6 +24,7 @@ app = dash.Dash()
 app.config.suppress_callback_exceptions = True
 
 app.webviz_settings = {
+    "shared_settings": webviz_config.SHARED_SETTINGS_SUBSCRIPTIONS.transformed_settings({{ shared_settings }}, {{ config_folder }}),
```
HansKallekleiv: black?

webviz_config/templates/webviz_template.py.jinja2
```diff
@@ -59,6 +59,7 @@ app.title = "{{ title }}"
 app.config.suppress_callback_exceptions = True
 
 app.webviz_settings = {
+    "shared_settings": webviz_config.SHARED_SETTINGS_SUBSCRIPTIONS.transformed_settings({{ shared_settings }}, {{ config_folder }}),
```
HansKallekleiv: black?

PR #165, opened 2019-12-05
Add theming for plotly figures

### LINKED ISSUE #164: Investigate usage of plotly.templates in webviz theme setting

HansKallekleiv:
Closes #164.


anders-kiaer:
@HansKallekleiv Should we do the following changes before merge?
1. Change `app.webviz_settings["plotly_theme"]` to `app.webviz_settings["theme"]`, and at the same time let it point to the actual theme instance.
2. Containers/plugins then needs to access is as `app.webviz_settings["theme"].plotly_theme`.

The benefit is then that we in the `Theme` class' `plotly_theme` getter-function can make sure a deep copy of the plotly theme is provided, making sure containers/plugins by mistake don't alter the original theme and doesn't need code for making sure to make a copy of the `plotly_theme` layout.

webviz_config/_theme_class.py
```diff
@@ -1,3 +1,6 @@
+from copy import deepcopy
```
anders-kiaer: Since we only use `deepcopy` once, consider using `import copy` here and then typing `copy.deepcopy()` further below where it is used.

PR #180, opened 2020-01-14
Allow io.BytesIO as webviz storage type

HansKallekleiv:


webviz_config/webviz_store.py
```diff
@@ -25,7 +26,7 @@ def register_function(self, func):
 
         return_type = inspect.getfullargspec(func).annotations["return"]
 
-        if return_type not in [pd.DataFrame, pathlib.Path]:
+        if return_type not in [pd.DataFrame, pathlib.Path, io.BytesIO]:
             raise NotImplementedError(
                 "Currently only storage of dataframes "
                 "and file resources are implemented."
```
anders-kiaer: Update error message. Maybe we can use this opportunity to move the expanding list `[pd.DataFrame, pathlib.Path, io.BytesIO]` to a constant, and then use that constant both in the `if` statement, and in the `NotImplementedError`.

webviz_config/webviz_store.py
```diff
@@ -175,6 +179,8 @@ def build_store(self):
                     output.to_parquet(f"{path}.parquet")
                 elif isinstance(output, pathlib.Path):
                     shutil.copy(output, f"{path}{output.suffix}")
+                elif isinstance(output, io.BytesIO):
+                    open(f"{path}", "wb").write(output.getvalue())
```
anders-kiaer: Remove `f`-string since the string `path` is used directly?

Perhaps use `pathlib.Path.write_bytes()` in order to make sure the file is properly closed.

webviz_config/webviz_store.py
```diff
@@ -140,6 +141,9 @@ def get_stored_data(self, func, *args, **kwargs):
                 return pd.read_parquet(f"{path}.parquet")
             if return_type == pathlib.Path:
                 return pathlib.Path(glob.glob(f"{path}*")[0])
+            if return_type == io.BytesIO:
+                with open(f"{path}", "rb") as filehandle:
```
anders-kiaer: Remove `f`-string since the string `path` is used directly?

webviz_config/webviz_store.py
```diff
@@ -140,6 +141,9 @@ def get_stored_data(self, func, *args, **kwargs):
...
+                    return io.BytesIO(filehandle.read())
```
anders-kiaer: `pathlib.Path.read_bytes()` would be more compact to use here perhaps.

PR #204, opened 2020-02-24
Add webviz pytest fixture

### LINKED ISSUE #203: Pytest webviz fixture
Dash provides their own pytest fixture, however, still using that fixture, testing different Webviz plugins requires repeating boilerplate code.

Provide a pytest fixture, based on e.g. Dash's `dash_duo` fixture, to faciliate testing of Webviz plugins.

Some features to consider:
- The Webviz fixture should itself be based on `dash_duo`.
- Should faciliate creating the `app` instance, and create an instance of the plugin in a Webviz setting (with given arguments).
- Should be possible to test with CSP settings enabled.
- Should be possible to check if selenium console logs are empty or not.
- Should be possible to check if file resources outside "white-listed" folders are used in portable mode (to ensure the plugin indeed works in a "portable" setting).
- Save screenshots at user defined stages through the test, and optionally compare with stored pictures at given location (visual regression/testing).
- Should be possible to automatically check if tour guide element ID's (if defined) actually exists.

anders-kiaer:
Closes #203 when finished.


anders-kiaer:
When done, also add an integration test for functionality added in #269.

PR #219, opened 2020-04-07
Deep update plotly layout

### LINKED ISSUE #215: ThemeClass: Deep merge user defined Plotly layout dict with theme layout
Plotly layout dicts often have lower level settings which should be possible to default without the risk of unfortunate overwriting in a `layout_dict.update(user_defined_layout_dict)`

asnyv:
Resolves #215

### DIFF
--- a/webviz_config/_theme_class.py
+++ b/webviz_config/_theme_class.py
@@ -75,6 +75,34 @@ class WebvizConfigTheme:
             else:
                 self._csp[key] = value
 
+    def create_themed_layout(self, layout: dict) -> dict:
+        """
+        Create a new Plotly layout dict by merging the input layout with the theme layout,
+        prioritizing the input layout if there are conflicts with the theme layout. In addition:
+        For the special case of multiple xaxes or yaxes, e.g. xaxis2 and xaxis3 (for a secondary
+        and tertiary xaxis), the axis will get the theme xaxis/yaxis layout, unless they are
+        defined themselves as e.g. xaxis2 in the theme layout. Note that e.g. xaxis2 still needs to
+        be in the input layout, just not in the theme layout.
+        """
+        # pylint: disable=too-many-nested-blocks
+        def deep_update(update: dict, ref: dict) -> dict:
+            for key, value in ref.items():
+                if key in update:
+                    if isinstance(value, dict):
+                        if isinstance(update[key], dict):
+                            update[key] = deep_update(update[key], value)
+                        if key in ["xaxis", "yaxis"]:
+                            for kkey in update:
+                                if kkey not in ref and kkey.startswith(key):
+                                    update[kkey] = deep_update(update[kkey], value)
+                else:
+                    update[key] = value
+            return update
+
+        return deep_update(
+            copy.deepcopy(layout), copy.deepcopy(self._plotly_theme["layout"])
+        )
+
     @property
     def csp(self) -> dict:
         """Returns the content security policy settings for the theme."""
--- a/tests/test_create_themed_layout.py
+++ b/tests/test_create_themed_layout.py
@@ -1,0 +1,148 @@
+from pathlib import Path
+
+import mock
+import dash
+
+from webviz_config import WebvizConfigTheme
+from webviz_config.common_cache import CACHE
+from webviz_config.themes import default_theme
+from webviz_config.plugins import _syntax_highlighter
+
+
+def test_create_themed_layout():
+    # Make an instance of the theme class
+    test_theme = WebvizConfigTheme(theme_name="test")
+    # Add a plotly theme to the new theme
+    test_theme.plotly_theme = {
+        "layout": {
+            "font": {"family": "Arial"},
+            "colorscale": {
+                "diverging": [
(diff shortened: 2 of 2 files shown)


tests/test_layout_updater.py
```diff
@@ -0,0 +1,37 @@
...
+            "colorway": ["#FFFFFF", "#0000FF"],
+            "xaxis": {"title": {"text": "Title"}, "type": "log", "constrain": "domain"},
+            "legend": {"font": {"family": "Arial"}},
+        }
+        # make a deep updated layout from specified and theme layouts
+        new_layout = default_theme.update_plotly_layout(specified_layout)
+        # test values
+        assert new_layout["colorway"] == specified_layout["colorway"]
+        assert new_layout["xaxis"]["title"]["text"] == "Title"
+        assert dash_duo.get_logs() == [], "browser console should contain no error"
```
anders-kiaer, line 37: Looks good. Would also be nice with a very simple test (can be in the same file as this), where:

1. Create a new theme, where a simple `plotly` layout dictionary is attached.
1. Make a simple dictionary variable as an example of a "plugins `plotly` layout modificitaion".
1. Call `update_plotly_layout` with this variable. The output is checked/verified to be the expected merged dictionary.
1. The `plotly` theme inside the Webviz theme should be checked that is unmodified after the call.
1. The variable from step two should be checked that is unmodified by the `update_plotly_layout` call.

webviz_config/_theme_class.py
```diff
@@ -75,6 +75,33 @@ def adjust_csp(self, dictionary: dict, append: bool = True) -> None:
             else:
                 self._csp[key] = value
 
+    def update_plotly_layout(self, layout: dict) -> dict:
```
anders-kiaer, line 78: I guess the name is inspired by current usage, e.g. `dict_a.update(dict_b)`, where `dict_a` is updated with the content of `dict_b`. Here we instead (I guess) want to merge two dictionaries (the plotly layout in the theme, and the plugin/user provided plotly layout), and return the new object? Should we use e.g. `merge`/`combine`/`...` instead of `update`?

asnyv: I would say that we are still updating `dict_a` (the theme layout) with the contents of `dict_b` (the layout for the specific case), where contents of `dict_b` is prioritized over `dict_a` if it exists in both (so that you can force deviations from the theme). `merge` and `combine` sounds more like they are two equally prioritized dicts during the process.

anders-kiaer: In addition we create a new dict/object I guess (i.e. not only mutate `dict_a` as in `dict_a.update`)?

asnyv: Aah, ok, now I see your concern. Sounds a bit like you update the `plotly theme layout`. What about something like `create_themed_layout`? Then it should be clear that we are making a completely new `dict`?

anders-kiaer: Good name suggestion 👍

webviz_config/_theme_class.py
```diff
@@ -75,6 +75,38 @@ def adjust_csp(self, dictionary: dict, append: bool = True) -> None:
             else:
                 self._csp[key] = value
 
+    def create_themed_layout(self, layout: dict) -> dict:
+        """
+        Create a new Plotly layout dict by merging the input layout with the theme layout,
+		prioritizing the input kayout if there are conflicts with the theme layout. In addition:
+		For the special case of multiple xaxes or yaxes, E.g. xaxis2 and xaxis3 (for a secondary
+		and tertiary xaxis), will get the theme xaxis/yaxis layout, unless they are defined
```
anders-kiaer, line 83: Missing a word after the first `,`?

webviz_config/_theme_class.py
```diff
@@ -75,6 +75,38 @@ def adjust_csp(self, dictionary: dict, append: bool = True) -> None:
...
+        themselves as e.g. xaxis2 in the theme layout. (Note that e.g. xaxis2 still needs to be
+		in the input layout, just not in the theme layout.
+        """
```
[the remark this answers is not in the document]
asnyv: Hmm, yes, seems like regular tabs instead of 4 spaces had found there way in somehow 🤔 Surprised black/pylint didn't catch something like that. Fixed now.

webviz_config/_theme_class.py
```diff
@@ -75,6 +75,37 @@ def adjust_csp(self, dictionary: dict, append: bool = True) -> None:
...
+                    if isinstance(value, dict):
+                        if isinstance(update[key], dict):
+                            update[key] = deep_update(update[key], value)
+                        if key in ["xaxis", "yaxis"]:
+                            for kkey in [
+                                nkey
+                                for nkey in update
+                                if nkey not in ref and nkey.startswith(key)
+                            ]:
+                                update[kkey] = deep_update(update[kkey], value)
```
anders-kiaer, line 100:
```suggestion
                            for kkey in update:
                                if kkey not in ref and kkey.startswith(key):
                                    update[kkey] = deep_update(update[kkey], value)
```

PR #225, opened 2020-04-24
Limit dash version to 1.10

HansKallekleiv:
Necessary due to https://github.com/plotly/dash/issues/1200 and problems with webviz-subsurface.


HansKallekleiv:
As this issue is limited to webviz-subsurface it will be fixed there.

PR #230, opened 2020-05-02
Use urls instead of tabs for main layout

HansKallekleiv:
An attempt to replace current tab based page navigation with [dcc.Location](https://dash.plotly.com/urls).

- Components will require persistence as they are reloaded on page switching
- Initial load of app should be must faster, but page switching might be somewhat slower

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [UNRELEASED] - YYYY-MM-DD
+
+### Changed
+
+- [#230](https://github.com/equinor/webviz-config/pull/230) - Instead of using
+`dcc.Tabs` to give the impression of a "multipage app", webviz now uses `dcc.Link` and
+`dcc.Location`. This has two main advantages: Big applications can have significantly
+faster loading time, as only callbacks on selected "page" fire initially. In addition,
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [UNRELEASED] - YYYY-MM-DD
+
+### Changed
+
+- [#230](https://github.com/equinor/webviz-config/pull/230) - Instead of using
+`dcc.Tabs` to give the impression of a "multi page app", webviz now uses `dcc.Link` and
```
asnyv, line 12: Alternatively multi-page (https://en.wiktionary.org/wiki/multipage)
```suggestion
`dcc.Tabs` to give the impression of a "multipage app", webviz now uses `dcc.Link` and
```

CHANGELOG.md
```diff
@@ -4,6 +4,20 @@ All notable changes to this project will be documented in this file.
...
+`dcc.Location`. This has two main advantages: Big applications can have significantly
+faster loading time, as only callbacks on selected "page" fire initially. In addition,
+the user can navigate with forward/backward buttons in browser, as well as getting
+an URL they can share with other in order to point them to the correct "page". Plugin
```
asnyv, line 16:
```suggestion
an URL they can share with others in order to point them to the correct "page". Plugin
```

asnyv (changes requested):
LGTM, only a couple of details in the changelog text 🥇

asnyv (approved):
LGTM 🎉

PR #233, opened 2020-05-18
Improve selectors in TablePlotter

HansKallekleiv:
Some minor fixes for TablePlotter:

- Use wcc.select for filters
- Make optional selectors (e.g. color, size) clearable
- Remove `y` variable from histogram

Also removed some css from default theme for wcc.select to work properly.


webviz_config/plugins/_table_plotter.py
```diff
@@ -26,6 +26,7 @@ class TablePlotter(WebvizPluginABC):
 * `csv_file`: Path to the csv file containing the tabular data. Either absolute
               path or relative to the configuration file.
 * `plot_options`: A dictionary of plot options to initialize the plot with
+* `filter_cols`: Dataframe columsn that can be used to filter data
```
anders-kiaer, line 29: Small typo.

PR #243, opened 2020-06-30
Semi-deterministic port in order to use persistence

### LINKED ISSUE #242: Dash persistence friendly port selection
Dash provides [persistence](https://dash.plotly.com/persistence) and a [`dcc.Store`](https://dash.plotly.com/dash-core-components/store) class that facilitates e.g. storing user chosen values across sessions. They both typically use [`window.localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage).

`localStorage` can only access stored values from same combination of `host:port`. Currently, on `localhost`, we ask for _some_ available port (typically returning a random port in the range `>= 1025` in order to easily allow for multiple `localhost` applications running at the same time).
https://github.com/equinor/webviz-config/blob/d691787f62700a5cda2e550fe7aec5bc1953713c/webviz_config/utils/_available_port.py#L13-L14

This probably needs to be more deterministic, e.g. start by checking port `5000` and increase `+1` from there, in order to get  benefit of persistence across sessions.

anders-kiaer:
Closes #242.

### DIFF
--- a/webviz_config/utils/_available_port.py
+++ b/webviz_config/utils/_available_port.py
@@ -1,18 +1,38 @@
 import os
 import socket
+from typing import Optional
 
 
-def get_available_port() -> int:
+def get_available_port(preferred_port: Optional[int] = None) -> int:
     """Finds an available port for use in webviz on localhost. If a reload process,
     it will reuse the same port as found in the parent process by using an inherited
     environment variable.
+
+    If preferred_port is given, ports in the range [preferred_port, preferred_port + 20)
+    will be tried first, before an OS provided random port is used as fallback.
     """
 
+    def is_available(port: int) -> bool:
+        with socket.socket() as sock:
+            try:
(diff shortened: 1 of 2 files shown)


webviz_config/utils/_available_port.py
```diff
@@ -1,18 +1,37 @@
...
-        sock = socket.socket()
-        sock.bind(("localhost", 0))
-        port = sock.getsockname()[1]
-        sock.close()
+        port = None
+
+        if preferred_port is not None:
+            for port_to_test in range(preferred_port, preferred_port + 20):
+                if port is None and is_available(port_to_test):
+                    port = port_to_test
```
asnyv, line 29: Could break loop when successful, but the max 20 iterations and `port is None` check makes the gain limited of course
```suggestion
                    port = port_to_test
                    break
```

PR #250, opened 2020-07-16
Allow skipping code in init function during portable build

### LINKED ISSUE #249: Expose instance of `WebvizRunMode` during build and runtime
Plugins may have code in their` __init__` method which isn't required to run when building portables. E.g. if their are methods called during `__init__` which are also included in `add_webvizstore`, the code will be executed twice during portable build.

For the examples in `webviz-config`, the gain is negligable, but when processing large datasets in `webvizstore`-decorated functions, the difference in build time may be significant.

asnyv:
Resolves #249 

This is a suggestion, might be a neater way to do this, but works at least. Plugin writer can define an (optional) input `_skip_portable_build` (can of course switch this to something else), a boolean that can be used to skip parts of the `__init__` function during `copy_data.py` when building portables. Modified the `ExamplePortable`. This should preferably be defaulted to False to keep backwards compatibility (otherwise old portable apps have to be rebuilt).

### DIFF
--- a/webviz_config/_config_parser.py
+++ b/webviz_config/_config_parser.py
@@ -14,7 +14,14 @@ from . import WebvizPluginABC
 from .utils import terminal_colors
 
 warnings.simplefilter("default", DeprecationWarning)
-SPECIAL_ARGS = ["self", "app", "container_settings", "_call_signature", "_imports"]
+SPECIAL_ARGS = [
+    "self",
+    "app",
+    "container_settings",
+    "_call_signature",
+    "_imports",
+    "_skip_portable_build",
+]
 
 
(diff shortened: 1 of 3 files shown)


anders-kiaer:
Could it be a possibility that we do not need to support this extra functionality after skipping initial callback globally in the app, and also follow-up with some best practice guidelines for plugin authors (i.e. that `__init__` should not contain time consuming code, but rather in an initial callback)?

anders-kiaer:
Closing this as we might have a way of solving this use case through https://github.com/equinor/webviz-config/issues/249#issuecomment-925696930 and the introduction of `WebvizRunMode`.

PR #262, opened 2020-08-13
Fix doc typos

anders-kiaer:


asnyv:
Should we also mention that the list is a list of argument combinations formatted as dictionaries?

HansKallekleiv (approved):
👍

PR #269, opened 2020-08-24
Change screenshot file name

DanSava:
Alow plugin users to change the filename used when the screenshot is saved.


anders-kiaer (approved):
Thanks @DanSava - LGTM! 👍

Issue #270, opened 2020-08-27
Code style black 20.8b1


asnyv:
Black 20.8b1 requires some minor changes for 🟢

PR #275, opened 2020-08-28
Allow indented plugin docstrings + Indent rest of markdown docstring

### LINKED ISSUE #272: Generated documentation does not support indented docstrings
If the docstring is indented, like
```python
class Markdown(WebvizPluginABC):
    """Renders and includes the content from a Markdown file.

    ---

    * **`markdown_file`:** Path to the markdown file to render and include. \
                       Either absolute path or relative to the configuration file.

    ---

    Images are supported, and should in the markdown file be given as either
    relative paths to the markdown file itself, or as absolute paths.

    > The markdown syntax for images has been extended to support \
      providing width and/or height for individual images (optional). \
      To specify the dimensions write e.g.
    > ```markdown
    > ![width=40%,height=300px](./example_banner.png "Some caption")
    > ```

    """
```
the "special section" with `>` is not rendered correctly.

Some plugins might want to consistently indent docstring lines. Remove all "consistent" indentation in docstring lines before giving it further to the documentation machinery?

asnyv:
Support for plugin docstrings that are indented to be used with `webviz docs` and `webviz schema`.
Also added an indent in `Markdown` plugin as the `black` indent in #270 had skipped the quoted section (this might be a risk going forward).

---

### Contributor checklist

- [X] :tada: This PR closes #272 
   - [X] Support indented docstrings.
   - [X] Indented section of `Markdown` docstring.
- [X] :robot: I have added tests, or extended existing tests, to cover any new features or bugs fixed in this PR.

### DIFF
--- a/tests/test_docstring.py
+++ b/tests/test_docstring.py
@@ -1,0 +1,76 @@
+from webviz_config._docs._build_docs import _split_docstring
+
+
+def test_split_docstring():
+    tests = [
+        (
+            "This is a test with only description.",
+            ["This is a test with only description."],
+        ),
+        (
+            (
+                " This is a test with only description, "
+                "but which has leading and trailing spaces.  "
+            ),
+            [
+                (
+                    "This is a test with only description, "
+                    "but which has leading and trailing spaces."
+                )
+            ],
+        ),
+        (
+            "Test with some newlines\n\n  \n    and a 4 space indent",
+            ["Test with some newlines\n\n\nand a 4 space indent"],
+        ),
+        (
+            "Test with a \n    4 space and a \n  2 space indent\n",
+            ["Test with a \n  4 space and a \n2 space indent"],
+        ),
+        (
+            "Test with a \n    4 space and a \n  2 space indent\n",
+            ["Test with a \n  4 space and a \n2 space indent"],
+        ),
+        (
+            (
+                " This is a test with description, arguments and data input."
+                "\n\n    ---\n\n    The docstring is indented by 4 spaces:\n    "
+                "* The argument list has a sub list indented by 8 spaces\n        like this."
+                "\n    ---\n    The data input section is not very interesting."
+            ),
+            [
+                "This is a test with description, arguments and data input.\n",
+                (
+                    "\nThe docstring is indented by 4 spaces:\n"
+                    "* The argument list has a sub list indented by 8 spaces\n    like this."
+                ),
+                "The data input section is not very interesting.",
+            ],
+        ),
+        (
+            (
+                "\tThis is a test with description, arguments and data input,"
+                " where indents are made with tabs and not spaces-"
+                "\n\n\t---\n\n\tThe docstring is indented by 1 tab:\n"
+                "\t* The argument list has a sub list indented by 2 tabs\n\t\tlike this."
+                "\n\t---\n\tThe data input section is not very interesting."
+            ),
+            [
+                (
+                    "This is a test with description, arguments and data input,"
+                    " where indents are made with tabs and not spaces-\n"
+                ),
+                (
+                    "\nThe docstring is indented by 1 tab:\n"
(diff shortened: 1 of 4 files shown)


webviz_config/_docs/_build_docs.py
```diff
@@ -162,3 +162,23 @@ def build_docs(build_directory: pathlib.Path) -> None:
...
+    indent_spaces = 1000  # far beyond max width
+    for line in lines[1:]:
+        if line.strip() == "":
+            continue
+        indent_spaces = min(indent_spaces, (len(line) - len(line.lstrip())))
+    unindented_lines = [lines[0]] + [
+        line.strip() if line.strip() == "" else line[indent_spaces:]
+        for line in lines[1:]
+    ]
+    return "\n".join(unindented_lines).split("\n---\n")
```
anders-kiaer, line 184: Does this work also with tab characters `\t`? It is not considered good Python style to have tabs instead of spaces, but some package authors might still use tabs? Something to support here?

webviz_config/_docs/_build_docs.py
```diff
@@ -162,3 +162,23 @@ def build_docs(build_directory: pathlib.Path) -> None:
...
     )
+
+
+def _split_docstring(docstring: str) -> List[str]:
+    """Divide docstring by splitting on ---, also unindent
+    in case of indented docstrings (similar to this one)
+    """
+    lines = docstring.strip().split("\n")
+    if len(lines) == 1:
+        return lines
```
anders-kiaer, line 173: Do we need this special casing? Rest of the code should run ok?
```python
a = [42]
# a[1:] == []
```

webviz_config/_docs/_build_docs.py
```diff
@@ -162,3 +162,23 @@ def build_docs(build_directory: pathlib.Path) -> None:
...
+
+    indent_spaces = 1000  # far beyond max width
+    for line in lines[1:]:
+        if line.strip() == "":
+            continue
+        indent_spaces = min(indent_spaces, (len(line) - len(line.lstrip())))
```
anders-kiaer, line 179:
```suggestion
indent_spaces = min([len(line) - len(line.lstrip()) for line in lines[1:] if line.strip() != ""])
```
Not tested, but something along these lines might be possible + easier to read?

tests/test_docstring.py
```diff
@@ -0,0 +1,52 @@
...
+            [
+                "This is a test with description, arguments and data input.\n",
+                (
+                    "\nThe docstring is indented by 4 spaces:\n"
+                    "* The argument list has a sub list indented by 8 spaces\n    like this."
+                ),
+                "The data input section is not very interesting.",
+            ],
+        ),
+    ]
```
anders-kiaer, line 50: Worth adding a test case where the final closing `"` is on its own line, potentially with no indentation?

webviz_config/_docs/_build_docs.py
asnyv: No, we don't 👍 Good catch 🎉

webviz_config/_docs/_build_docs.py
asnyv: Think strip() without arguments should remove any form of whitespace, including tabs, but can add a test for it to be certain

webviz_config/_docs/_build_docs.py
asnyv: Edit, with the next suggested change it was needed a similar handling. Changed to try-except.

webviz_config/_docs/_build_docs.py
anders-kiaer: Good. Mix is ok not to support 😄

asnyv: Have also added a test with tabs 👍

anders-kiaer (approved):
🚀👍

PR #279, opened 2020-09-04
Scrollable sidebar

HansKallekleiv:
### Contributor checklist

- [x] :scroll: I have broken down my PR into the following tasks:
   - [x] scrollable sidebar
   - [x] fix logo

scroll


anders-kiaer:
Does it work Firefox as well? (Some [technical reference](https://developer.mozilla.org/en-US/docs/Web/CSS/Webkit_Extensions#Supported_in_Firefox_with_-webkit-_prefix))

anders-kiaer (approved):
LGTM 🎨

PR #289, opened 2020-09-10
Use new pip resolver

anders-kiaer:
~As for now,~ this is a check of the new pip resolver. ~We will merge this when the new resolver actually becomes the default (in October, and with the release of `pip==20.3`).~


anders-kiaer:
`pip==20.3` has passed planned release date by 18 days, still not released. Suggest we add this, and also add a information box in readme stating that people using `pip < 20.3` (which anyway would be quite some also after new `pip` release should enable the new resolver).

HansKallekleiv (approved):
👍

PR #295, opened 2020-09-15
Specify menu width

### LINKED ISSUE #291: Menu disappears if plugin content is broad (e.g. Dash `DataTables`)
Regression bug after #279.

### LINKED ISSUE #301: Limit width of menu on the left hand side
**Is your feature request related to a problem? Please describe.**
If users write long page titles, the menu can be quite wide. Would be better to have more line breaks and leave more room for the plugin visualization.

**Describe the solution you'd like**
A max width for the menu.

**Describe alternatives you've considered**
Possibility for forced line breaks with e.g. `\n` using `white-space: pre-wrap;` in [div.styledButton.tab](https://github.com/equinor/webviz-config/blob/623ab6fbabf85dd658be1d668d9f6e216939fdd2/webviz_config/static/assets/webviz_layout.css#L11-L27)

HansKallekleiv:

---

### Contributor checklist

- [x] :tada: This PR closes #291 and closes #301.
- [x] :scroll: I have broken down my PR into the following tasks:
   - [x] Overflow page content 
   - [x] Tested on existing webviz subsurface plugins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,12 @@ enforced by `webviz-config` (inline script hashes are added automatically).
 - [#281](https://github.com/equinor/webviz-config/pull/281) - Now uses `importlib` instead of `pkg_resources` for
 detecting plugin entry points and package versions.
 
+### Fixed
+- [#295](https://github.com/equinor/webviz-config/pull/295) - Menu width is now specified. This both ensures
+the menu does not collapse if plugin content is wide, as well as not too wide itself if page titles are long
+(instead page titles in the menu are wrapped). Also added `<meta name="viewport" content="width=device-width, initial-scale=1">`
+for better experience on mobiles.
+
 ## [0.1.2] - 2020-09-09
 ### Added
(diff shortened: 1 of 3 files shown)


anders-kiaer:
Seeing #301, are we better of having min+max width specified on the menu?

It is also a bit unclear to me how we want to solve overflow-x on individual dashboards/plugins, and if that would be in conflict with the current addition in this PR. E.g. the parallell coordinates plugin is typically very vide, and the user will have to scroll all the way to the bottom to scroll it horizontally. Might want to have overflow scrollbar on the plot/plugin itself? Maybe we should try to solve both at the same time?

HansKallekleiv:
Agreed. min/max width on sidebar is better.

anders-kiaer:
As discussed in sprint meeting today, @HansKallekleiv will try with `width` only and some `vw%` (+ verify that breakpoint functionality still make us support narrow displays).

anders-kiaer (approved):
LGTM! :+1:

PR #307, opened 2020-09-21
Report Komodo release used

### LINKED ISSUE #305: komodo version
Internally in Equinor, https://github.com/equinor/komodo is used to deploy software. If we detect that komodo is used, should we add a comment in generated portable app which komodo release it was built with?

anders-kiaer:
Closes #305.

### DIFF
--- a/webviz_config/_build_webviz.py
+++ b/webviz_config/_build_webviz.py
@@ -1,3 +1,4 @@
+import os
 import sys
 import time
 import shutil
@@ -39,6 +40,8 @@ def build_webviz(args: argparse.Namespace) -> None:
     (build_directory / "theme_settings.json").write_text(
         installed_themes[args.theme].to_json()
     )
+
+    _store_komodo_release(build_directory)
 
     try:
         if args.portable:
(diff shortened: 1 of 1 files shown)


webviz_config/_build_webviz.py
```diff
@@ -133,3 +136,21 @@ def run_webviz(args: argparse.Namespace, build_directory: pathlib.Path) -> None:
                 f"{terminal_colors.END}"
             )
             raise excep
+
+
+def _store_komodo_release(build_directory: pathlib.Path) -> None:
+    """Internally in Equinor, https://github.com/equinor/komodo is one platform used
+    for software deployment. For portable builds we want, until cloud hosting is common for
+    everyone, mention for the user which specific Komodo release (~= pip list) that was used
```
asnyv, line 144:
```suggestion
    everyone, to mention for the user which specific Komodo release (~= pip list) that was used
```

webviz_config/_build_webviz.py
```diff
@@ -133,3 +136,21 @@ def run_webviz(args: argparse.Namespace, build_directory: pathlib.Path) -> None:
...
+    to build the portable Webviz application.
+
+    This function stores the Komodo release used, if Komodo is used, in a file komodo_release.txt.
+    """
+
+    if "KOMODO_RELEASE" in os.environ and sys.executable.startswith("/prog/res/komodo"):
+        (build_directory / "komodo_release.txt").write_text(
+            "This portable Webviz application was built with Komodo release:\n"
+            f"{os.environ['KOMODO_RELEASE']}\n\n"
+            "Try sourcing this Komodo release if you "
```
asnyv, line 154: Suggest that we explicitly state the required command to source the komodo release (probably best to use the csh alternative)

asnyv (changes requested):
Think we should consider to be even more explicit on what/how to source the correct environment.
Most users will understand what is currently there, but doesn't hurt to add the full command.

anders-kiaer:
Will be moved to https://github.com/equinor/webviz-config-equinor, as it is primarily Equinor that uses Komodo.

PR #308, opened 2020-09-21
Resolve UnclosedFile+ResourceWarning + graceful exit on KeyboardInterrupt

### LINKED ISSUE #298: Configuration file is unclosed
https://github.com/equinor/webviz-config/blob/623ab6fbabf85dd658be1d668d9f6e216939fdd2/webviz_config/_config_parser.py#L152

### LINKED ISSUE #299: Wait for subprocess to stop
We currently do `app_process.kill()` in [`_build_webviz.py`](https://github.com/equinor/webviz-config/blob/master/webviz_config/_build_webviz.py), but we don't follow up with `app_process.wait()` which we should do.

Could at the same time move `time.sleep(1)` into `try/except` and catch `KeyboardInterrupt` as well, such that the user is not presented with a traceback in an expected even such that `KeyboardInterrupt` (instead the user could be presented with a good bye message or similar).

anders-kiaer:
---

### Contributor checklist

- [X] :tada: This PR closes #298 and closes #299.
   - [X] Use `pathlib.Path.read_text()` instead of `with/open` context.
   - [X] Follow best practice and wait for subprocess to stop.
   - [X] Graceful exit on `KeyboardInterrupt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ detecting plugin entry points and package versions.
 feedback quicker if something is wrong.
 
 ### Fixed
+- [#308](https://github.com/equinor/webviz-config/pull/308) - Fixed `UnclosedFile` and `Resource` warnings,
+which would appear if developer enabled e.g. `export PYTHONWARNINGS=default`. Also, Webviz now gracefully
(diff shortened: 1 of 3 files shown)


HansKallekleiv (approved):
lgtm 👍

PR #310, opened 2020-09-22
Give RuntimeWarning if cached/stored functions use pandas object arguments

### LINKED ISSUE #211: pandas objects wrt. flask-caching and webviz_store
Flask caching with objects as input introduces a risk of not capturing changes in the object.
Read the warning under [memoization in the Flask documentation](https://flask-caching.readthedocs.io/en/latest/#memoization). Flask only checks the `repr()` of the object in the input, which might not by unique. The issue has been encountered with Pandas in Webviz, but can occur with any object instance that doesn't have a unique `repr()` output. 

Temporary solution is **to avoid using other input types than basic Python types** for `@CACHE.memoize` decorated methods. E.g. converting DataFrames to dict.

anders-kiaer:
Closes #211.


HansKallekleiv (approved):
👍

PR #313, opened 2020-09-23
min-width on menu + let menu expand on small screens

anders-kiaer:
The menu still can "collapse" wrt. width if content is wide. Therefore adding a `min-width`.

Also, on small screens, the Dash provided `Tabs` have a break point at `800px`. We only want to specify menu width at screens larger than `800px` such that on small screens we utilize the available space.


HansKallekleiv (approved):
Much better 👍

PR #318, opened 2020-09-30
Improved Docker and cloud support 

### LINKED ISSUE #150: Install Python packages in Dockerfile
For third-party webviz plugins, the users should be provided an easy way of getting also those Python packages into the generated Dockerfile.

anders-kiaer:
Closes #150.

- [X] Remove dependency on webviz base image.
- [X] Use same Python slim base image as user had when creating the portable app.
- [X] Install same webviz framework version as user has locally.
- [X] Install same plugin projects, with same version, as user has locally.
- [X] Create Radix config automatically.


CONTRIBUTING.md
```diff
@@ -584,3 +533,40 @@ $$\alpha = \frac{\beta}{\gamma}$$
...
 Example of auto-built documentation for `webviz-config` can be seen
 [here on github](https://equinor.github.io/webviz-config/).
+
+## Make your plugin project available
+
+It is strongly recommended to store your plugin project code base in a `git` solution,
+e.g. [in a new GitHub repository](https://github.com/new). If possible, it is also
+_highly_ recommended to let it be an open source repository. This has several advantages:
+ - Others can reuse your work - and help each other achieve better code quality, add more features and share maintenance
+ - You will from the beginning make sure you keep a good separation between visualization code and raw data to be visualized.
```
HansKallekleiv, line 543: replace `and raw data to be visualized` with e.g. `and data loading/processing code`

CONTRIBUTING.md
```diff
@@ -584,3 +533,40 @@ $$\alpha = \frac{\beta}{\gamma}$$
...
+ - It becomes easier for your plugin users to use it in e.g. Docker and in cloud hosting.
+ - ...
```
HansKallekleiv, line 545: remove

HansKallekleiv (dismissed):
Looks good!
I am not entirely sure that the `radix configuration` should be included in `webviz-config` as this is Equinor specific? Could instead the `webviz-config-equinor` be extended ? Something we can come back to. 
Besides that only a minor comment to the documentation :)

webviz_config/templates/radixconfig.yaml.jinja2
```diff
@@ -0,0 +1,66 @@
...
+          port: 5000
+      environmentConfig:
+        - environment: prod
+          resources:
+            requests:
+              memory: "400Mi"
+              cpu: "100m"
+            limits:
+              memory: "4G"
+              cpu: "1000m"
```
asnyv, line 59: Is it ideal to default these settings? 
I guess we can override them after initial setup, but we've seen significantly higher memory usage on localhost than what has been defaulted here?
Also: I would think that such a spread between requested and maximum memory usage could lead to out of memory errors if you need the limit value, and you're a bit unlucky?

anders-kiaer: I think it is outside scope in this PR to intelligently find good values based on what the user has added to the app. But we should review the default values before merge :+1: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/ and https://www.radix.equinor.com/docs/reference-radix-config/#resources give some information.

HansKallekleiv (approved):

PR #320, opened 2020-10-06
Use getvalue instead of seek+read

anders-kiaer:
`io.BytesIO()` has a short hand for `seek(0) + read()`, which is `getvalue()`.


asnyv (approved):
LGTM 👍

PR #321, opened 2020-10-06
Support plotly.js screenshot download

### LINKED ISSUE #317: Allowing blob: in img-src
- [x] Check if there are security aspects to this
- [x] If none or small, add `blob:` to `img-src` CSP.

Alternative solution:
- Check `plotly.js` "Download to png" source code and see if they could/should use the same approach as in `webviz-core-components` (which also uses `blob:`, but instead of attaching it to an `img` element it let the user download it directly).

anders-kiaer:
`plotly.js` "Download screenshot to png" requires `blob:` in `img-src` CSP.

---

### Contributor checklist

- [X] :tada: This PR closes #317.


asnyv (approved):
✔🙈

PR #324, opened 2020-10-08
Move standard plugins to entrypoints

anders-kiaer:
This is for two reasons:
- [x] We demonstrate how pure plugin projects should do it, and do it on equal terms.
- [x] Some soon-to-arive utility functions within `webviz-config` become easier since we don't need to special case the plugins coming from config (all plugins will come from entrypoints).


asnyv (approved):
👍

PR #325, opened 2020-10-09
Extract plugin distribution information on import

### LINKED ISSUE #323: Currently assume top level package name equals distribution package name
Even though it quite often that what the package/distribution is called in `setup.py` is the same as what you `import`, e.g.
`pip install x` and `import x` is the same `x`, they could in principle be different.

E.g. `pip install libecl` make you able to do `import ecl` (and not `import libecl`).

The current implicit assumption would lead to `importlib.metadata.version` to fail in cases where distname `!=` toplevel package name for a plugin package.

anders-kiaer:
Closes #323.

This change enables us to automatically include plugin packages in Docker/cloud deployment, based on what the user have asked for in config.

Separately, it could also enable us to in plugin wrapper toolbar optionally later add buttons for link to plugin source/issue tracker, documentation...


HansKallekleiv (approved):
LGTM 👍

PR #333, opened 2020-10-23
Raise a warning if multiple plugins exist with same name

### LINKED ISSUE #292: Raise a warning if multiple plugins exist with same name
Since we are a plugin framework, there might be plugin projects creating plugins with the same name.

Raise a warning during runtime to give attention to name conflict. This could be done in
https://github.com/equinor/webviz-config/blob/623ab6fbabf85dd658be1d668d9f6e216939fdd2/webviz_config/plugins/__init__.py#L37-L39
by simply checking if a plugin with name is already in `__all__`. 

The warning should probably include package name of both plugins in order to make it easy for the user to understand where they come from (one way of getting package name from the plugin reference is to use `inspect` from std.lib).

JoakimJohansen:
##hacktoberfestivus

Gives warning if multiple plugins exist with same name in different packages, along with the conflicting package names.

Contributors: The GreenTeam w/
 torgeha
 JoakimJohansen
 antoneskov
---

### Contributor checklist

- [x] :tada: This PR closes #292.


antoneskov:
@anders-kiaer , can we please have a review, evt. approval by the end of  "hacktoberfestivus"?

webviz_config/plugins/plugin_utils.py
```diff
@@ -0,0 +1,27 @@
+import warnings
+
+def write_metadata(distributions, metadata):
```
anders-kiaer, line 3: This function we might want to give another, since it also loads the plugins (even though it at the same time collect metadata about them).

I also _think_ that it is the module having the `globals()` calls defined in it (i.e. this module in this case) that will get them in their namespace, and not modules importing the function and calling it. Regardless it makes it difficult to understand what is available in which namespace. Maybe the best and most explicit would be that this function also returns references to the loaded plugins, and then `plugins/__init__.py` which calls this function, is responsible for putting the plugins in its namespace.

webviz_config/plugins/plugin_utils.py
```diff
@@ -0,0 +1,27 @@
...
+                
+                metadata[entry_point.name] = {
+                    "dist_name": dist.metadata["name"],
+                    "dist_version": dist.version,
+                    "documentation_url": project_urls.get("Documentation"),
+                    "download_url": project_urls.get("Download"),
+                    "tracker_url": project_urls.get("Tracker"),
+                }
+                    
+                globals()[entry_point.name] = ep
```
anders-kiaer, line 27: Add newline

webviz_config/plugins/plugin_utils.py
```diff
@@ -0,0 +1,27 @@
...
+        for entry_point in dist.entry_points:
+            if entry_point.group == "webviz_config_plugins":
+                project_urls = {
+                    value.split(",")[0]: value.split(",")[1].strip()
+                    for (key, value) in dist.metadata.items()
+                    if key == "Project-URL"
+                }
+
+                ep = entry_point.load()
+                if entry_point.name in globals():
```
anders-kiaer, line 14:
```suggestion
                if entry_point.name in metadata:

```

anders-kiaer (commented):
@antoneskov, @JoakimJohansen et al. - thanks for the PR 🎉 👏 Love the tests. 🤖 

We will get this in (if that is enough feedback wrt. "hacktoberfestivus"), but would need some small changes before ready for merge to `master`.

Added some feedback as comments below. Also CI requires running [`black`](https://github.com/psf/black) on the changed files to make it happy 🕵️‍♂️ If you don't have time I can also do the necessary changes, but that might be first next week.

JoakimJohansen:
Requested changes has been made @anders-kiaer.

anders-kiaer:
@asnyv and/or @HansKallekleiv: :green_circle: lights. This is ready for final review now from one of you.

asnyv (approved):
LGTM 🎉

PR #336, opened 2020-11-09
Remove unused css code

HansKallekleiv:


anders-kiaer (commented):
What do we want to call the collection of components/elements we visually group together with this `css` class? Maybe we can then use that name instead of the generic `frame` in the class name (which could also indicate that it is the current frame around the complete plugin layout)?

anders-kiaer:
Are we quite certain that we end up with supporting this instead from the `webviz-core-components` point of view? If so we limit the scope of this PR to CSS cleanup and get it in?

HansKallekleiv:
> @HansKallekleiv Are we quite certain that we end up with supporting this instead from the `webviz-core-components` point of view? If so we limit the scope of this PR to CSS cleanup and get it in?

Yes, I think that's the best solution. Providing this as a layout component instead of a css class will make more sense for most plugin authors.
I will clean up the pr.

anders-kiaer (dismissed):

anders-kiaer (approved):

Issue #14, opened 2019-05-07
Support for sub pages


HansKallekleiv:
A user wants to have sub pages to organize a lot of content.

- [ ] Add support for tabs in the config file

- [ ] Add tabs in the jinja templating

- [ ] Add theming of tabs

HansKallekleiv:
One solution can be to have collapsible elements on the main sidebar.
Collapsible side bar from `dash-bootstrap-components` :
https://github.com/facultyai/dash-bootstrap-components/tree/master/examples/multi-page-apps/sidebar-with-submenus
![image (1)](https://user-images.githubusercontent.com/16436291/74968395-0dc3b000-541b-11ea-996f-2fc08f63e9c3.png)

PR #339, opened 2020-11-16
Not rely on page called 'Front page'

anders-kiaer:
After #230, generated app started using Dash `dcc.Location` and `dcc.Link` instead of tabs for multi-page apps. This enabled URL routing. At the same time, the root url `/` special case was set to show "Front page", however, users did not from before necessarily have a page called "Front page", as the "Front page" shown earlier was simply the first page in the list.

Change back to behavior prior to #230 - i.e. show first page in config as "Front page" and not rely on page title. We can revisit ideal wanted behaviour when solving #14 and deprecate config syntax accordingly.


HansKallekleiv (approved):
Tested. Looks good 👍

Issue #332, opened 2020-10-23
localhost https


anders-kiaer:
As `webviz` is using SSL localhost (`https://`), it appears to be a challenge with Chrome in the sense that if the user later go to something non-https `localhost` link, Chrome in many cases will redirect the user to `https` (based on previos visits to a `https` visited localhost page).

If any mitigating actions can be done on our side, we should try to aim for the browser not redirecting to https on later visits to localhost by default.

**Informative SO thread:**
https://stackoverflow.com/questions/25277457/google-chrome-redirecting-localhost-to-https

anders-kiaer:
Some potentially relevant references:
* https://github.com/equinor/webviz-config/blob/ca9c273bb382e6c0639fc9b4e03ccf2cb0e004eb/webviz_config/templates/webviz_template.py.jinja2#L58
* https://github.com/GoogleCloudPlatform/flask-talisman

PR #345, opened 2020-11-25
Add oauth2 authorization code flow support

thezultimate:

---

### Contributor checklist

   - [ ] Task 1
   - [ ] Task 2

### DIFF
--- a/.github/workflows/webviz-config.yml
+++ b/.github/workflows/webviz-config.yml
@@ -44,6 +44,7 @@ jobs:
 
       - name: 📦 Install webviz-config with dependencies
         run: |
+          pip install urllib3==1.24.3
           pip install 'pandas==1.1.4' # Pinned to 1.1.4 due to pandas 1.1.5 bug with pylint: https://github.com/pandas-dev/pandas/issues/38355
           pip install --upgrade pip==20.2.4 # Pinned to 20.2.4 for black install compatibility: https://github.com/psf/black/issues/1847
           pip install .
--- a/.github/workflows/webviz-config.yml
+++ b/.github/workflows/webviz-config.yml
@@ -44,7 +44,6 @@ jobs:
 
       - name: 📦 Install webviz-config with dependencies
         run: |
-          pip install urllib3==1.24.3
           pip install 'pandas==1.1.4' # Pinned to 1.1.4 due to pandas 1.1.5 bug with pylint: https://github.com/pandas-dev/pandas/issues/38355
           pip install --upgrade pip==20.2.4 # Pinned to 20.2.4 for black install compatibility: https://github.com/psf/black/issues/1847
           pip install .
@@ -65,7 +64,7 @@ jobs:
           black --check webviz_config tests setup.py
           pylint webviz_config tests setup.py
           bandit -r -c ./bandit.yml webviz_config tests setup.py
-          # mypy --package webviz_config --ignore-missing-imports --disallow-untyped-defs --show-error-codes
+          mypy --package webviz_config --ignore-missing-imports --disallow-untyped-defs --show-error-codes
 
       - name: 🤖 Run tests
         run: |
--- a/.github/workflows/webviz-config.yml
+++ b/.github/workflows/webviz-config.yml
@@ -65,7 +65,7 @@ jobs:
           black --check webviz_config tests setup.py
           pylint webviz_config tests setup.py
           bandit -r -c ./bandit.yml webviz_config tests setup.py
-          mypy --package webviz_config --ignore-missing-imports --disallow-untyped-defs --show-error-codes
+          # mypy --package webviz_config --ignore-missing-imports --disallow-untyped-defs --show-error-codes
 
       - name: 🤖 Run tests
         run: |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [UNRELEASED] - YYYY-MM-DD
 
 ### Changed
+- [#345](https://github.com/equinor/webviz-config/pull/345) - Added Oauth2 
+Authorization Code flow support for Azure AD applications.
+
 - [#374](https://github.com/equinor/webviz-config/pull/374) - Removed Webviz
   SSL certificate generation and forcing of HTTPS connections.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [UNRELEASED] - YYYY-MM-DD
 
-### Changed
+### Added
 - [#345](https://github.com/equinor/webviz-config/pull/345) - Added Oauth2 
 Authorization Code flow support for Azure AD applications.
 
+### Changed
 - [#374](https://github.com/equinor/webviz-config/pull/374) - Removed Webviz
   SSL certificate generation and forcing of HTTPS connections.
 
--- a/setup.py
+++ b/setup.py
@@ -80,6 +80,7 @@ setup(
         "flask-talisman>=0.6",
         "jinja2>=2.10",
         "markdown>=3.0",
+        "msal>=1.5.0",
         "pandas>=1.0",
         "pyarrow>=0.16",
         "pyyaml>=5.1",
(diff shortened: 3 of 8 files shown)


anders-kiaer (commented):
This is great @thezultimate. 🎉👍🥳

One thing I've been thinking about: Should we remove the `--oauth` command line argument? Instead I have a suggestion we can discuss going along something like this:

- Plugins (which are Python classes) can choose to have a property method `oauth2(self)` (or some other descriptive name.
- If `>=1` plugin have that property method AND it retuns `True` during app startup, we do what you know do when `--ouauth2` is provided.

Being a plugin property method, the plugins will also be able to give `True` /  `False` based on user configuration input in the YAML file (e.g. the plugin deduces that  if user has not specified files on disk, it is expected to use external API etc... the logic here will be plugin specific).

We can take a short video session and discuss? I'm also open for contributing myself in the same PR here in order to move it to 🥅 .

thezultimate:
> This is great @thezultimate. 🎉👍🥳
> 
> One thing I've been thinking about: Should we remove the `--oauth` command line argument? Instead I have a suggestion we can discuss going along something like this:
> 
> * Plugins (which are Python classes) can choose to have a `@property` method `def oauth2(self):` (or some other descriptive name.
> * In the `jinja2` generated app script, if `>=1` plugin instances have that property method defined AND it returns `True` during app startup, we do what you know do when `--oauth2` is provided.
> 
> Being a plugin property method, the plugins will also be able to give `True` / `False` based on user configuration input in the YAML file (e.g. the plugin deduces that if user has not specified files on disk, it is expected to use external API etc... the logic here will be plugin specific).
> 
> We can take a short video session and discuss? I'm also open for contributing myself in the same PR here in order to move it to 🥅 .

I think I have implemented the changes you suggested in the last commit. Would it be possible for you to check? Or someone else if you are not available? Thanks.

webviz_config/templates/webviz_template.py.jinja2
```diff
@@ -113,6 +113,21 @@ else:
             ])]),
         html.Div(className="pageContent", id="page-content")])
 
+plugins = []
+{% for page in pages %}
+{% for content in page.content %}
+{% if content is not string %}
+plugins.append(webviz_config.plugins.{{ content._call_signature[0] }})
```
anders-kiaer, line 120: I think we want to create the `plugins` list prior to the for loop on line 73 (initializing plugins can be time consuming, depending on plugin logic, and it would be nice if line 124 and 79 could be using the same initialized plugins).

@asnyv can probably help you out if you want more details.

webviz_config/templates/webviz_template.py.jinja2
```diff
@@ -113,6 +113,21 @@ else:
...
+{% endif %}
+{% endfor %}
+{% endfor %}
+use_oauth2 = any(plugin.oauth2 for plugin in plugins if hasattr(plugin, "oauth2"))
+oauth2 = None
+if use_oauth2:
+    oauth2 = webviz_config.Oauth2(app.server)
+    oauth2.set_oauth2_endpoints()
+    if __name__ != "__main__":
+        oauth2.set_oauth2_before_request_decorator()
```
anders-kiaer, line 130:
```suggestion
oauth2 = webviz_config.Oauth2(app.server) if use_oauth2 else None
```
given that `set_oauth2_endpoints()` becomes part of `Ouaht2.__init__` (does that make sense, i.e. do you want to always call it on initizalition?) and that 
```python
oauth2.set_oauth2_before_request_decorator()
```
is moved to an `else:` block after the already existing `if __name__ == "__main__"` block.

setup.py
```diff
@@ -87,6 +87,7 @@ def get_long_description() -> str:
         "importlib-metadata>=1.7; python_version<'3.8'",
         "typing-extensions>=3.7; python_version<'3.8'",
         "webviz-core-components>=0.1.0",
+        "msal>=1.5.0",
```
anders-kiaer, line 90: Move this to alphabetically correct position?

anders-kiaer (changes requested):
Nice refactoring @thezultimate ! 🎉 Now we are close on getting this into `master`. 

Some comments below.

It would also be nice with a new paragraph in https://github.com/equinor/webviz-config/blob/master/CONTRIBUTING.md explaining this nice new optional feature for plugin authors🚀

webviz_config/_oauth2.py
```diff
@@ -0,0 +1,171 @@
...
+        and its expiration date is created. This flask session object can be
+        accessed from Webviz plugin.
+
+        To get the access token: flask.session.get("access_token")
+        To get the expiration date: flask.session.get("expiration_date")
+
+        An Azure AD application should be registered, and the following environment
+        variables should be set: TENANT_ID, CLIENT_ID, CLIENT_SECRET, SCOPE.
+
+        SESSION_SECRET_KEY environment variable is also needed for signing the
```
asnyv, line 52: Update comments these with the `WEBVIZ_` prefix

webviz_config/_oauth2.py
```diff
@@ -0,0 +1,171 @@
...
+def get_login_uri(url_root: str) -> str:
+    return replace_http_with_https(url_root + "login")
+
+
+def get_auth_redirect_uri(url_root: str) -> str:
+    return replace_http_with_https(url_root + "auth-return")
+
+
+def replace_http_with_https(url: str) -> str:
+    return url.replace("http://", "https://")
```
asnyv, line 171: Webviz on `localhost` will probably soon go from `https` to `http`. See [webviz-config/issues/332](https://github.com/equinor/webviz-config/issues/332) and [webviz-ert/issues/87](https://github.com/equinor/webviz-ert/issues/87).

I guess this might create an issue with this solution when running on `localhost` in the future, or?

webviz_config/templates/webviz_template.py.jinja2
```diff
@@ -151,3 +155,6 @@ if __name__ == "__main__":
         dev_tools_silence_routes_logging=False,
       {% endif %}
     )
+else:
+    if use_oauth2:
```
asnyv, line 159: add a comment that this is only necessary when not running on localhost, as it is implemented as part of the `LocalhostToken`?

.github/workflows/webviz-config.yml
```diff
@@ -44,6 +44,7 @@ jobs:
 
       - name: 📦 Install webviz-config with dependencies
         run: |
+          pip install urllib3==1.24.3
```
asnyv, line 47: Is pinning urllib3==1.24.3 necessary?

thezultimate: Yes, otherwise the build will break. It think this was suggested by @anders-kiaer .

webviz_config/_oauth2.py
thezultimate: The redirect URI for Azure AD app may be fine. But some logics should be changed (small change). We can wait until `localhost` is moved from `https` to `http` if you think this is the best route.

HansKallekleiv (approved):
LGTM 👍

webviz_config/_oauth2.py
```diff
@@ -0,0 +1,162 @@
...
+
+        To get the access token: flask.session.get("access_token")
+        To get the expiration date: flask.session.get("expiration_date")
+
+        An Azure AD application should be registered, and the following environment
+        variables should be set: WEBVIZ_TENANT_ID, WEBVIZ_CLIENT_ID,
+        WEBVIZ_CLIENT_SECRET, WEBVIZ_SCOPE.
+
+        WEBVIZ_SESSION_SECRET_KEY environment variable is also needed for signing the
+        session cookie.
```
asnyv, line 48: Replaced by autogenerated key, isn't it?

thezultimate: Ah yes, this should be removed :)

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [UNRELEASED] - YYYY-MM-DD
 
 ### Changed
+- [#345](https://github.com/equinor/webviz-config/pull/345) - Added Oauth2 
+Authorization Code flow support for Azure AD applications.
+
```
asnyv, line 12:
```suggestion
### Added
- [#345](https://github.com/equinor/webviz-config/pull/345) - Added Oauth2 
Authorization Code flow support for Azure AD applications.

### Changed
```

PR #367, opened 2020-12-09
Added py.typed marker file to announce that package supports typing

### LINKED ISSUE #366: Make type hints available to package consumers
The `webviz-config` package already has inline type hints but currently these are not available to package consumers. For example, the `webviz-subsurface` package, which has recently gained type hints will not be able to see the type hints from `webviz-config`. This means that running mypy linting on `webviz-subsurface` will not detect wrong usage of types imported from `webviz-config` since all imported types will be of type `Any`.

According to [PEP 561](https://www.python.org/dev/peps/pep-0561/) we need to add a `py.typed` marker file to our package and ensure that this file gets installed with the package.

sigurdp:
Added py.typed marker file as specified in  [PEP 561](https://www.python.org/dev/peps/pep-0561/) to announce that this package supports typing.

---

### Contributor checklist

- [X] :tada: This PR closes #366.


webviz_config/py.typed
```diff
@@ -0,0 +1 @@
+# Marker file to indicate that this package supports typing, ref PEP 561
```
anders-kiaer, line 1: Add newline?

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [UNRELEASED] - YYYY-MM-DD
 
+### Changed
+- [#367](https://github.com/equinor/webviz-config/pull/367) - Made type information
+available to package consumers by indicating support for typing as specified in PEP 561.
```
anders-kiaer, line 11:
```suggestion
available to package consumers by indicating support for typing as specified in [PEP 561](https://www.python.org/dev/peps/pep-0561/).
```

anders-kiaer (approved):
Looks good to me @sigurdp! 🚀 Some very minor optional suggestions.

PR #368, opened 2020-12-10
Make special (global) webviz_settings argument available to plugins

### LINKED ISSUE #352: Too magical app object
Plugin authors can request access to the `app` object (i.e. the Dash app instance). This is used in order to be able to attach callbacks etc.

However, other webviz specific structures are also accessible as `app.webviz_settings` at the moment.

Split into two special arguments? I.e. `app` and `webviz_settings`? `app` is then a pure Dash app without any additional properties, and `webviz_settings` is Webviz specific data/information.

sigurdp:
Instead of piggybacking webviz_settings as an attribute onto the Dash application object, make (global) Webviz settings available to plugins through special argument `webviz_settings.` This argument is an instance of the `WebvizSettings` class.

This PR also deprecates access to webviz_settings as an attribute on the Dash application instance object that is currently being passed to the plugins as the special `app` argument. Also note that the new `WebvizSettings` class does not contain the `portable` attribute.

To start using the new scheme, and avoid the deprecated functionality, you will have to modify your plugin's `__init__()` function to include the special `webviz_settings` argument. E.g. change the signature from:
<pre>
def __init__(self, app: dash.Dash, csv_file: Path) -> None:
</pre>
to
<pre>
def __init__(self, app: dash.Dash, <b>webviz_settings: WebvizSettings</b>, csv_file: Path) -> None:
</pre>

And for accessing the settings you will have to replace code such as this:

```
the_shared_settings = app.webviz_settings["shared_settings"]
the_theme = app.webviz_settings["theme"]
```
with
```
the_shared_settings = webviz_settings.shared_settings
the_theme = webviz_settings.theme
```

Todos:
- [x] Update documentation
- [x] Update change log, including deprecation
- [x] Spawn new issue to refactor plugins in webviz-subsurface

---

### Contributor checklist

- [x] :tada: This PR closes #352 
   - [ ] Task 1
   - [ ] Task 2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [UNRELEASED] - YYYY-MM-DD
 
 ### Changed
+- [#368](https://github.com/equinor/webviz-config/pull/368) - Made Webviz global
+settings available to plugin implementations through special `webviz_settings` 
+argument. This argument is an instance of the `WebvizSettings` class and currently
+contains both the `shared_settings` and `theme` properties.
 - [#367](https://github.com/equinor/webviz-config/pull/367) - Made type information
-available to package consumers by indicating support for typing as specified in [PEP 561](https://www.python.org/dev/peps/pep-0561/).
+available to package consumers by indicating support for typing as specified in 
+[PEP 561](https://www.python.org/dev/peps/pep-0561/).
+
+### Deprecated
+- [#368](https://github.com/equinor/webviz-config/pull/368) - Access to `webviz_settings`
+as an attribute on the Dash application instance object (currently being passed to the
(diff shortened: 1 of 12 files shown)


sigurdp:
This is a first cut implementation with several issues to be discussed:
* Shape and implementation style for WebvizSettings class
  - Location and file name of implementation
  - Properties in WebvizSettings should probably be immutable since they get passed on to all plugins?
  - Should we try and enforce this through type hints and/or by returning deep copies through properties?
* Current implementation adds a `portable` attribute to the Dash app object
  - What is the use case for this attribute? Cannot find any references to it.
  - Also poses a consistency problem - what value should it have when portable app is being *generated*?
* Regarding deprecation of functionality
  - Is this how we want to do it?
  - Also location and file name of implementation
  - Visibilty of deprecation warnings?

HansKallekleiv (approved):
LGTM 👍

PR #373, opened 2021-01-06
Fix for import of TypedDict for Python versions older than 3.8

sigurdp:
The current way of importing TypedDict (for type hinting) using try-except seems to confuse mypy and causes errors during type checking. See for example in: https://github.com/equinor/webviz-subsurface/runs/1647251891?check_suite_focus=true

This PR fixes the issue by checking against Python version instead of using try-except during import.

See: 
https://mypy.readthedocs.io/en/stable/common_issues.html#python-version-and-system-platform-checks
and: 
https://github.com/python/typeshed/issues/3500#issuecomment-558398738

Also updated CHANGELOG to reflect the already published release 0.2.5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [UNRELEASED] - YYYY-MM-DD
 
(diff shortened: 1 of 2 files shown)


asnyv (approved):
LGTM 👍

Issue #288, opened 2020-09-10
Webviz launches Chromium when my OS default is Firefox


dotfloat:
When I do `webviz build [...]`, Webviz launcher the Chromium browser, when my OS default browser is Firefox. This happens due to a hardcoded preference for Google Chrome and Chromium:

https://github.com/equinor/webviz-config/blob/61f0fb16c377f3f821983a4e1a986cf92585d364/webviz_config/utils/_localhost_open_browser.py#L59-L63

Consider removing these lines so that my OS default browser is used rather than Chromium.

(To set default browser on GNOME: go to Settings -> Search "Default Applications" -> Select "Chromium" as your Web Browser)

anders-kiaer:
Thanks @dotfloat for the report. :tada: :slightly_smiling_face: 

**History:** The preference for Chrome/Chromium has historically been to make `webviz` easier to use for internal non-technical first-time users with the combination RHEL6 + KDE (where most people would have an ancient version of Konqueror as default without knowing).

Users having a preference to e.g. Firefox can set [preferred browser](https://equinor.github.io/webviz-config/#/?id=user-preferences).

**Going forward:** All users have recently moved away from RHEL6 :tada: And people are encouraged to use GNOME. Webviz has also "settled" a bit in the FMU community (= easier for users to ask questions if something goes wrong). So could consider now removing preferred browsers and go with default OS browser. 

The tricky part is there are still users on RHEL7 + KDE, which I think by default gets Konquerer version 4 (from 2008!). So the best solution would perhaps be to preferably use OS default browser, but give a notice/warning if OS default is unsupported/old. However, I haven't found an easy portable way of getting browser version :thinking: Browser name looks easy though:
```python
>>> import webbrowser
>>> 
>>> webbrowser.get().name
'firefox'
>>> 
```

So might be that the best possible solution is to use OS default, and communicate with/help users that don't know they have a _very_ old default browser? :thinking:

<sup>Or we can [do as Google simply give a warning if Konqueror is used](https://9to5google.com/2019/12/16/google-blocking-obscure-browsers-gmail/) :laughing: </sup>

dotfloat:
Thanks for the write-up. I sympathise with the problem, although it's it seems to be a fix at the wrong level. We should bribe Knut Petter to have Firefox/Chromium be the default on KDE, even if KDE is technically unsupported in Equinor.

[...]

PR #374, opened 2021-01-12
Remove SSL cert for localhost

### LINKED ISSUE #332: localhost https

dotfloat:
This PR removes SSL certification creation and usage when running webviz in localhost mode. This makes it HTTP-only for localhost with no HTTPS option.

It isn't complete in the sense that we still have to consider all cases. To my knowledge, the industry standard within web applications is to have the application (webviz) only do things using HTTP, and let a reverse proxy deal with HTTPS. I'm not entirely sure how this applies to Azure, but I imagine a similar thing is true there. That means that our use of `flask-talisman` (the security hardening addon for Flask) should perhaps not be as strict.

This should also be documented somehow. The workflow of our users depends on there existing certificates.

Resolves #332
Resolves https://github.com/equinor/webviz-ert/issues/87


webviz_config/templates/webviz_template.py.jinja2
```diff
@@ -67,7 +66,9 @@ app._deprecated_webviz_settings = {
 CACHE.init_app(server)
 
 theme.adjust_csp({"script-src": app.csp_hashes()}, append=True)
-Talisman(server, content_security_policy=theme.csp, feature_policy=theme.feature_policy)
+if __name__ != "__main__":
+    # Do not increase security when running this script directly
+    Talisman(server, content_security_policy=theme.csp, feature_policy=theme.feature_policy)
```
asnyv, line 71: We would prefer to keep Talisman for the csp settings. As linked in #332 there are additional settings options here: https://github.com/GoogleCloudPlatform/flask-talisman some of which seems related to http/https/hsht. 

Also, did you check if you [got this issue in Chrome]( https://stackoverflow.com/q/25277457) after making the change? (That chrome redirects you to https since you have already been on https localhost before?

asnyv: I tested the following (might be that the opposite if-else is nicer (== and not !=) 
```
if __name__ != "__main__":
    Talisman(server, content_security_policy=theme.csp, feature_policy=theme.feature_policy)
else:
    # Do not increase security when running this script directly
    Talisman(server, content_security_policy=theme.csp, feature_policy=theme.feature_policy, strict_transport_security=False, force_https=False)
```
This only seems to work in Chrome if I have already run one app with something like this first though (setting a 10 second time-out on the strict_transport_security instead of the default 1 year):
```
Talisman(server, content_security_policy=theme.csp, feature_policy=theme.feature_policy, strict_transport_security_max_age=10)
```
Will have to check a bit more to figure out if we can somehow avoid this (since we want the full change in one release, preferably without all Chrome users getting issues...)

anders-kiaer: > Will have to check a bit more to figure out if we can somehow avoid this (since we want the full change in one release, preferably without all Chrome users getting issues...)

Not sure if there is anything we can do :thinking:  My current understanding now is that HSTS (after first visit) further prevents man-in-middle attacks by having a remembered header/setting saying that "on this page you should redirect to/use `https` as long as within the timeout".

@asnyv I guess preferably existing users should not need to manually go in and remove HSTS for `localhost` in their browser settings when we remove `https` `localhost`. Maybe we ASAP should add a very short HSTS localhost timeout (in a separate ~one line PR) and get it into `komodo`, and then merge this PR and release in e.g. the next `komodo` release? It is a balance between as little trouble for existing `webviz` users, and also try to quickly support what `webviz-ert` needs.

---

Of course the best "deprecation phase" period would be to have `flask` configured such that it both supports `http` and `https`, but don't enforce `https`. However, doesn't look like it is out of the box. I quickly tried
```python
from flask import Flask

app = Flask(__name__)

@app.route("/")
def hello():
    return "Hello World!"

if __name__ == "__main__":
    app.run(ssl_context='adhoc')
```
and that failed when I went to `http` :thinking:

asnyv: @anders-kiaer Not sure if this is really a solution. Many users are not running new apps every month, so they will not get this "fix". In addition, some users go back to older `komodo` versions to run older portables, and if they then step back, they will get in trouble. This will problem will also occur in the future if someone runs older portables with old versions of `webviz-config` / `komodo`. Fear it will haunt us for a long time...

anders-kiaer: Temporary entrypoint `webviz remove-hsts` which just uses existing certificates and creates an empty page "HSTS removed. You are ready for future versions!"?

dotfloat: We could change the port number.

Having this HSTS-disabler code would have to dig deep into Google Chrome implementation which is really far beyond the scope of webviz. Also consider that in order to be able to redirect `https` back to `http` you'll still need to have a valid certificate, and so you'll still need to have a step creating the cert before you can run Webviz.

A big concern of mine is how Webviz' scope seems to include Equinor's RGS environment, including Komodo. I believe this HSTS-disabler script, if required, should live outside of Webviz. Komodo has a way of executing scripts at enable-time, but this already seems to be a massively over-engineered solution.

I think the best solution in this is to:
1. Have `webviz certificate` tell the user that it no longer exists and exit.
2. When launching the localhost browser, tell the user in console what to do if they get a cert error (ie, direct them to some page in the webviz docs)

anders-kiaer: > We could change the port number.

That would be the preferred way, but I think HSTS ignores port. 🤔 [Reference](https://serverfault.com/questions/882026/hsts-on-main-port-80-not-on-other-ports). Can also be checked by doing:
1) Starting a `flask` (or `webviz`) instance with `https`, and visit the `localhost` page (the `https` certificate needs to be trusted by the browser).
2) Start a simple standard `http` flask server, with some other port `x` than in the previous step, and go to `http://localhost:x`. You should be redirected to `https`.

FWIW, `127.0.0.1` does not seem to use the "cached header" for `localhost`.

> Having this HSTS-disabler code would have to dig deep into Google Chrome implementation

Not sure if I understand, this is not Chrome specific. E.g. a temporary `webviz remove-hsts` entrypoint, which removes/overwrites a HSTS cached setting with `Talisman` default value, could roughly be something simple like
```python
import sys

from flask import Flask
from flask_talisman import Talisman

from webviz_config.certificate import LocalhostCertificate

app = Flask(__name__)

@app.route("/")
def hello():
    return "HSTS successfully removed! You are ready for this and further versions of webviz-config"

Talisman(app, strict_transport_security_max_age=0)

if __name__ == "__main__":
    app.run(host="localhost", ssl_context=LocalhostCertificate().ssl_context)
```
This should work regardless of browser. 🌍 

> A big concern of mine is how Webviz' scope seems to include Equinor's RGS environment, including Komodo. I believe this HSTS-disabler script, if required, should live outside of Webviz.

The issue here I would say is independent of RGS (#288 is influenced by RGS, but that is an independent task). Existing HSTS cached header setting would be relevant for all existing `webviz` users (having used a valid `localhost` certificate).

I like your idea with a console warning/info @dotfloat. 💡 Maybe we could consider something like this (iterating further on your suggestion @dotfloat):
1) We change `webviz certificate` to do nothing, except tell the user that the command has no effect and is deprecated (as new users don't need to create certificates).
2) We create a new flag `webviz certificate --remove` which does ~the HSTS overwriting thing above, and maybe ask user to remove previously user generated webviz certificates on disk (if they exist in the webviz user settings folder).
2) Change to `http` on generated `webviz` apps (as already done in this PR). If user certificates still exists on disk, write a warning on startup and refer to `webviz certificate --remove`.

dotfloat: So since this _is_ essentially a Google Chrome problem, there are two ways webviz could remove HSTS:
1. Know about Chrome to an extent and headlessly remove HSTS for localhost (if at all possible). This has the benefit of potentially being entirely invisible to the user.
2. Keep creating a certificate in order to provide an endpoint which removes HSTS, but visibly.

Number two is what you were suggesting, and it has the unfortunate problem in that we're requiring the user to perform an action, and we'll need to keep the cert code around for a long time in case someone a 11 months from now decides to open up webviz.

It definitely is a divide between Equinor/RGS and everyone else as far as I see:
* **Equinor/RGS** users are "web developers" running webviz in localhost
* **Everyone else** uses webviz as hosted on Azure or similar. This is the general way people use web services and fits well into eg. Chrome's way of operations.

While it definitely is not fair for our researchers to have to think about web development issues, I really am against solving what is a people problem with software engineering. It's easier to give the user a warning and not spend more development or brain time on it.

Regarding `webviz certificate --remove`. I'm not in favour:
1. The certificates take up a insignificant amount of space
2. If the user goes back to an earlier webviz they'll have to recreate it for no good reason
3. We don't expect users to know what a certificate is or why they really need it. Adding the ability to remove it will only confuse them even more and I would not expect the majority of people to actually use this
4. Those who are informed enough about the certificates and don't want it are able to figure out how to remove it themselves

dotfloat: I'll add that my general motto is "Less Code Is Gooder Than More Code", which is something I might find out is incorrect down the line.

From my POV, Webviz is an extension to Dash/Plot.ly which adds the ability to configure websites with a YAML file. What happens when running localhost on Google Chrome is outside of the goal and therefore should have as few lines of code dedicated to fixing it, preferably 0.

dotfloat:
The current changes are:
- Talisman only has forced-HTTPS disabled
- `webviz certificate` is a dummy command that `sys.exit(0)` (gracefully and will continue to work in scripts)
- Added documentation about HSTS in the introduction.md
- When starting a localhost browser, inform the user about potential problems on HSTS

asnyv:
Is it safe to have force-https disabled also when not running `localhost` @dotfloat? 🤔

dotfloat:
In my experience, the web application runs behind a reverse proxy. `nginx` or Apache `httpd` if you're hosting it yourself. On Azure, I think the correct way to deploy applications is by having them with no protocol-level security and let the platform deal with the rest. Ie, Azure should be the one to force HTTPS, not the application.

My reasoning is that I know that you can specify/buy SSL keys on Azure, and then it doesn't make sense to have the webapp deal with HTTPS, but Azure magically know how to inject SSL keys into the black box that is your web program.

I might be very wrong though, but I reckon this is something that needs to be brought up with security.

FWIW, all of my web development experience comes from using Ruby on Rails, which could've been forcing HTTPS in a smart but transparent way that I've never noticed.

webviz_config/command_line.py
```diff
@@ -175,3 +175,14 @@ def entrypoint_schema(args: argparse.Namespace) -> None:
     args = parser.parse_args()
 
     args.func(args)
+
+
+def _dummy_create_ca() -> None:
```
asnyv, line 180: The set_defaults sends the arguments to the function, currently fails.
```suggestion
def _dummy_create_ca(_: argparse.Namespace) -> None:
```

webviz_config/_localhost_token.py
```diff
@@ -28,6 +27,7 @@ class LocalhostToken:
 
     The port is used as a postfix on the cookie name in order to make sure that
     two different localhost applications running simultaneously do not interfere.
+
```
asnyv, line 30: Remove?
```suggestion
```

asnyv:
One thing to fix since `webviz certificate` commands crashes due to too many arguments given to `_dummy_create_ca`. 
Otherwise it looks good to me @dotfloat 😊🎉

dotfloat:
Fixed. `webviz certificate --auto-install` and co works as intended. :+1:

asnyv (approved):
LGTM 🚀 Thansk a lot @dotfloat 🎉👍

PR #377, opened 2021-01-14
Improved interface for discovery of and access to installed Webviz plugins

sigurdp:
First cut suggestion for possible refactor of the interface for discovery and access to all installed Webviz plugins.

This is not a finsihed solution, but currently meant as a basis for discussion and feedback on whether this is a sensible direction to proceed in. 

The suggestion implements access to the installed plugins through a "module based" singleton that wraps access to the underlying module level globals. This is implemented in `webviz_config/plugin_registry/__init__.py ` The prototype implementation discovers plugins and populates the module level variables when `plugin_registry` is imported, but this could instead be done lazily.


webviz_config/plugin_registry/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+from .._plugin_abc import WebvizPluginABC
+from ._load_webviz_plugins_with_metadata import load_webviz_plugins_with_metadata
+from ._load_webviz_plugins_with_metadata import PluginDistInfo
+
+try:
+    # Python 3.8+
+    from importlib.metadata import distributions  # type: ignore
+except ModuleNotFoundError:
+    # Python < 3.8
+    from importlib_metadata import distributions  # type: ignore
```
asnyv, line 15: This will not give similar issues as what caused #373?

webviz_config/plugin_registry/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+    return _all_installed_plugins[plugin_name]
+
+
+def plugin_metadata(plugin_name: str) -> PluginDistInfo:
+    return _all_installed_metadata[plugin_name]
+
+
+# Something along these lines will probably be needed for validation using pydantic
+def overloaded_init_methods_for_plugin(_plugin_name: str) -> List[Callable]:
+    return []
```
asnyv, line 40: Raise a `NotImplementedError` for now in case someone would try to use it? Or just remove and include later in pydantic PR?

webviz_config/plugin_registry/_load_webviz_plugins_with_metadata.py
```diff
@@ -56,4 +56,8 @@ def load_webviz_plugins_with_metadata(
 
                 loaded_plugins[entry_point.name] = entry_point.load()
 
+                # We should add a check here to verify that the loaded entry point
+                # is actually a WebvizPlugin. Something akint to the code in _get_webviz_plugins()
+                # which may then be deleted
```
asnyv, line 61: Yep we should 👍

webviz_config/_docs/_build_docs.py
```diff
@@ -105,9 +104,9 @@ def get_plugin_documentation() -> defaultdict:
     """
 
     plugin_doc = [
-        _document_plugin(plugin)
-        for plugin in _get_webviz_plugins(webviz_config.plugins)
-        if not plugin[0].startswith("Example")
+        _document_plugin((plugin_name, plugin_registry.plugin_class(plugin_name)))
+        for plugin_name in plugin_registry.all_plugin_names()
+        if not plugin_name.startswith("Example")
```
asnyv, line 109: Must say: I wasn't aware of this, and not anything to do with the PR, but to skip plugins that starts with `Example` in the doc is quite hacky and magic 😋

asnyv (commented):
Couple of small comments, but I think the concept looks good. 
Agree that some more meaningful attribute names than `_call_signature[x]` would be nice

webviz_config/_docs/_build_docs.py
anders-kiaer: Maybe we (in a separate PR) can consider #381 and then remove `if not plugin_name.startswith("Example")` here afterwards?

PR #380, opened 2021-01-21
Skip multiple version of plugin warning if the metadata is equal

### LINKED ISSUE #379: symlinked lib and lib64 in venv gives warning of multiple versions of same plugin
**Describe the bug**
Warnings written of multiple occurences of same plugin if `lib` and `lib64` are symlinked (appears like the package is installed twice for `importlib_metadata`).

**How to reproduce**
1. Create a virtual environment (using `python -m venv yourvenv`, the behavior seems different for `virtualenv`. Not sure why)
2. `pip install webviz-config`
3. `webviz build examples/basic_example.yaml`
4. See warnings in terminal

**Expected behavior**
Not write warning

**Additional context**
Suggest to ignore if the metadata is completely equal for the two occurences of the same entry point.

asnyv:
### Contributor checklist

- [X] :tada: This PR closes #379

### DIFF
--- a/webviz_config/plugins/_utils.py
+++ b/webviz_config/plugins/_utils.py
@@ -39,6 +39,18 @@ def load_webviz_plugins_with_metadata(
                 }
 
                 if entry_point.name in metadata:
+                    if metadata[entry_point.name] == PluginDistInfo(
+                        {
+                            "dist_name": dist.metadata["name"],
+                            "dist_version": dist.version,
+                            "documentation_url": project_urls.get("Documentation"),
+                            "download_url": project_urls.get("Download"),
+                            "tracker_url": project_urls.get("Tracker"),
+                        }
+                    ):
+                        # May occur if folders in virtual environment are symlinked.
+                        continue
+
                     warnings.warn(
                         f"Multiple versions of plugin with name {entry_point.name}. "
                         f"Already loaded from project {metadata[entry_point.name]['dist_name']}. "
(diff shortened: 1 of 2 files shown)


webviz_config/plugins/_utils.py
```diff
@@ -39,20 +39,34 @@ def load_webviz_plugins_with_metadata(
...
-                }
+                metadata[entry_point.name] = PluginDistInfo(
+                    {
+                        "dist_name": dist.metadata["name"],
+                        "dist_version": dist.version,
+                        "documentation_url": project_urls.get("Documentation"),
+                        "download_url": project_urls.get("Download"),
+                        "tracker_url": project_urls.get("Tracker"),
+                    }
+                )
```
anders-kiaer, line 69: Define this metadata dictionary once as a variable? Here it is introduced two times (more work to maintain)?

anders-kiaer (changes requested):
Looks like a reasonable change👍 

Maybe not the most common use case, but do we know what happens e.g. if:
1) `webviz-config` + `webviz-subsurface` is installed in system Python.
2) You create a venv with `--system-site-packages`.
3) You source the venv and `pip install --upgrade webviz-config webviz-subsurface`.

Will it currently give a warning (and then also after this one, if the `--upgrade` bumps the plugin package version)?

webviz_config/plugins/_utils.py
sigurdp: I suggest just leaving the duplication for now. My upcoming and revised https://github.com/equinor/webviz-config/pull/377 ends up rewriting/moving this code so the duplication can removed then.

HansKallekleiv (approved):

PR #384, opened 2021-01-23
Fix MyPy 0.800 errors in Markdown plugin

### LINKED ISSUE #385: Release of MyPy version 0.800 gives type checking errors in CI
New `Mypy` errors in `CI` after MyPy 0.800 was released:

```
webviz_config/generic_plugins/_markdown.py:37: error: Return type "Tuple[Any, ...]" of "handleMatch" incompatible with return type "Union[Tuple[Element, int, int], Tuple[None, None, None]]" in supertype "InlineProcessor"  [override]
webviz_config/generic_plugins/_markdown.py:37: error: Signature of "handleMatch" incompatible with supertype "Pattern"  [override]
webviz_config/generic_plugins/_markdown.py:46: error: Item "None" of "Optional[str]" has no attribute "startswith"  [union-attr]
webviz_config/generic_plugins/_markdown.py:52: error: Argument 1 to "Path" has incompatible type "Optional[str]"; expected "Union[str, _PathLike[str]]"  [arg-type]
webviz_config/generic_plugins/_markdown.py:60: error: Item "None" of "Optional[str]" has no attribute "split"  [union-attr]
Found 5 errors in 1 file (checked 45 source files)
```

asnyv:
### Contributor checklist

- [X] :tada: This PR closes #385.
   - [X] Decided to override incompatible return type in for error:
`    webviz_config/generic_plugins/_markdown.py:37: error: Return type "Tuple[Any, ...]" of "handleMatch" incompatible with return type "Union[Tuple[Element, int, int], Tuple[None, None, None]]" in supertype "InlineProcessor"  [override]`
As I would have to import another package for the `Element` class + the new type hint would be very long. Seems like we are not anymore allowed a courser type hint (in this case `tuple`) than in parent classes you override.
   - [X] The other errors were fixed by handling the possibility of `None` returned from the getters.

Has limited user interest, so skipped it from `CHANGELOG`. Tell me if you disagree.

### DIFF
--- a/webviz_config/generic_plugins/_markdown.py
+++ b/webviz_config/generic_plugins/_markdown.py
@@ -34,7 +34,7 @@ class _MarkdownImageProcessor(ImageInlineProcessor):
 
         super().__init__(image_link_re, md)
 
-    def handleMatch(self, m, data: str) -> tuple:  # type: ignore[no-untyped-def]
+    def handleMatch(self, m, data: str) -> tuple:  # type: ignore[no-untyped-def, override]
         image, start, index = super().handleMatch(m, data)
 
         if image is None or not image.get("title"):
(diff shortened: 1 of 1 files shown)


anders-kiaer (approved):
👍

PR #387, opened 2021-01-24
Allow plugin modules to register new webviz store return types

HansKallekleiv:

---

### Contributor checklist

   - [ ] Task 1
   - [ ] Task 2


anders-kiaer (commented):
Enabling plugins to register custom storage formats I guess is worth an extra feature 👍 (There are quite some domain specific objects/classes out there with their own storage/load functions).

The only "downside" I can see is that we then expose how we access `WebvizStorage` files to plugin authors, so maybe we should settle first on how we _think_ files will be accessed locally and in ☁️? Even better perhaps would be if there is an API suggestion introduced in this PR which doesn't limit ourselves to files available through OS file system (but e.g. support blob streaming using web API behind the scenes), but that abstraction is perhaps difficult.. 🤔

PR #416, opened 2021-03-31
Highlight selected page

HansKallekleiv:
Highlight selected button in sidebar.

---

### Contributor checklist

   - [ ] Add a callback that sets the className to `selectedButton` for the selected button.


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [UNRELEASED] - YYYY-MM-DD
+- [#416](https://github.com/equinor/webviz-config/pull/416) - Highlight selected
+button in main sidebar.
```
anders-kiaer, line 9:
```suggestion
page button in main sidebar.
```

webviz_config/templates/webviz_template.py.jinja2
```diff
@@ -117,12 +119,8 @@ else:
                     href="/",
                     {%- else -%}
                     "{{page.title}}",
-                    # We will create a webviz-core-components
-                    # component instead of styling dcc.Link's,
-                    # then we can more easily change className to
-                    # selectedButton for current page.
                     className="styledButton",
-                    id="{{page.id}}",
+                    id={"id":"{{page.id}}","type":"page_child"},
```
anders-kiaer, line 123:
```suggestion
                    id={"id": "{{page.id}}", "type": "page_child"},
```
In order to reduce risk of name conflicts with potential plugin pattern-matching callbacks, should we increase the `type` name detail level? `webviz_config_main_sidebar_link`?

tests/test_portable.py
```diff
@@ -1,7 +1,28 @@
 import sys
+import json
 import subprocess  # nosec
 
 
+def _stringify_object_id(uuid) -> str:
+    """Object ids must be sorted and converted to
+    css strings to be recognized as dom elements"""
```
anders-kiaer, line 8:
```suggestion
    CSS strings to be recognized as DOM elements"""
```

webviz_config/templates/webviz_template.py.jinja2
```diff
@@ -141,6 +139,23 @@ def update_page(pathname):
...
+@app.callback(
+    Output({"id": ALL, "type": "page_child"}, "className"),
+    Input("url", "pathname"),
+    State({"id": ALL, "type": "page_child"}, "href"),
+)
+def _update_class(pathname, children_hrefs):
+    """Update style to 'selectedButton' for the active page
+    else use 'styledButton'"""
+    if not children_hrefs:
+        raise PreventUpdate
```
anders-kiaer, line 152: When will this occur?

anders-kiaer (commented):
Initial review. :slightly_smiling_face: 

Is it correct that the Dash generated IDs of the menu buttons becomes strings of the format `'{\"id\"\:\"some_id\"\,\"type\"\:\"page_child\"\}'`? If so I guess downstream test pattern would need to change as well, in cases where you ["click" through pages with `selenium`](https://github.com/equinor/webviz-subsurface/blob/f5f482484011bcde29125464d766290f30161d1c/tests/integration_tests/test_raw_data_example.py#L56-L63). And that would imply the `type` string becomes public property, or better we provide a utility function for giving the Dash generated page IDs.

We can also discuss if we can prioritize https://github.com/equinor/webviz-core-components/issues/77 and solve highlighting through that one.

HansKallekleiv:
> Is it correct that the Dash generated IDs of the menu buttons becomes strings of the format `'{\"id\"\:\"some_id\"\,\"type\"\:\"page_child\"\}'`? If so I guess downstream test pattern would need to change as well, in cases where you ["click" through pages with `selenium`](https://github.com/equinor/webviz-subsurface/blob/f5f482484011bcde29125464d766290f30161d1c/tests/integration_tests/test_raw_data_example.py#L56-L63). And that would imply the `type` string becomes public property, or better we provide a utility function for giving the Dash generated page IDs.

Yes, thats correct , [this](https://github.com/plotly/dash/blob/dev/tests/integration/callbacks/test_wildcards.py#L266) is the only example I've found from Plotly. A utility function is a good idea as this conversion is needed for any selenium test against object IDs.

webviz_config/templates/webviz_template.py.jinja2
HansKallekleiv: never ;)

anders-kiaer:
> A utility function is a good idea as this conversion is needed for any selenium test against object IDs.

Maybe we should make a PR towards `plotly/dash` on this 🤔 I guess there are many Dash based projects wanting to e.g. test against (with `selenium`) or using client side callbacks. This is apparently where they do it today: https://github.com/plotly/dash/blob/597dc22c86de97530fb9b0a9bd9f7876bd0d7601/dash-renderer/src/actions/dependencies.js#L111-L123

anders-kiaer:
Just discovered that there is already a utility function for this in `dash`:
```python
>>> from dash.dash import stringify_id
>>> 
>>> stringify_id({"id": "some_id", "type": "some_type"})
'{"id":"some_id","type":"some_type"}'
>>> 
```

anders-kiaer:
Superseded by https://github.com/equinor/webviz-core-components/issues/77 which is in progress.

PR #426, opened 2021-04-25
Added centralized factory registry for sharing factories between plugins

sigurdp:
Added centralized/global `WebvizFactoryRegistry` that allows factories to be registered and shared between plugins. 
The class also facilitates storage of optional factory settings that are read from the YAML config file for later consumption when
an actual factory gets created and added to the registry.

The registry is merely managed and initialized by `webviz-config` and then exposed globally through `WEBVIZ_FACTORY_REGISTRY.`  Actual creation/registration of factories and subsequent retrival and use of factories will be done by the concrete factory classes and/or the plugins that utilize the factories. 

Note that currently, factory settings are not exposed publicly. Initial usage of these settings is assumed to be for experimental and internal tweaking purposes. The factory settings can be specified in the configuration file using the `internal_factory_settings` key, but the interpretation and usage of of the contained dictionary is left to the individual factory. A hypothetical section specifying factory settings could look something like this:

```yaml
internal_factory_settings:
  EnsembleTableProviderFactory:
    backing_implementation: arrow
    arrow_options:
      keep_files_open: yes
  EnsambleSummaryProviderFactory:
    backing_implementation: parquet
    parquet_options:
      keep_files_open: yes
      downcast_to_float: true
```

---

### Contributor checklist

### DIFF
--- a/webviz_config/templates/webviz_template.py.jinja2
+++ b/webviz_config/templates/webviz_template.py.jinja2
@@ -84,12 +84,7 @@ WEBVIZ_ASSETS.portable = {{ portable }}
 
 run_mode = WebvizRunMode.PORTABLE if {{portable}} else WebvizRunMode.NON_PORTABLE
 app_instance_info = WebvizInstanceInfo(run_mode, storage_folder)
-{% if internal_factory_settings is defined %}
-factory_settings_dict = {{ internal_factory_settings }}
-{% else %}
-factory_settings_dict = None
-{% endif %}
-WEBVIZ_FACTORY_REGISTRY.initialize(app_instance_info, factory_settings_dict)
+WEBVIZ_FACTORY_REGISTRY.initialize(app_instance_info, {{ internal_factory_settings if internal_factory_settings is defined else None }})
 
 use_oauth2 = False
 
--- a/webviz_config/templates/webviz_template.py.jinja2
+++ b/webviz_config/templates/webviz_template.py.jinja2
@@ -145,6 +145,8 @@ else:
             ])]),
         html.Div(className="pageContent", id="page-content")])
 
+WEBVIZ_FACTORY_REGISTRY.cleanup_resources_after_plugin_init()
+
 oauth2 = webviz_config.Oauth2(app.server) if use_oauth2 else None
 
 @app.callback(dash.dependencies.Output("page-content", "children"),
--- a/webviz_config/templates/webviz_template.py.jinja2
+++ b/webviz_config/templates/webviz_template.py.jinja2
@@ -20,8 +20,7 @@ from webviz_config.themes import installed_themes
 from webviz_config.common_cache import CACHE
 from webviz_config.webviz_store import WEBVIZ_STORAGE
 from webviz_config.webviz_assets import WEBVIZ_ASSETS
-from webviz_config.webviz_instance_info import WebvizRunMode
-from webviz_config.webviz_instance_info import WebvizInstanceInfo
+from webviz_config.webviz_instance_info import WebvizRunMode, WebvizInstanceInfo
 from webviz_config.webviz_factory_registry import WEBVIZ_FACTORY_REGISTRY
 from webviz_config.utils import deprecate_webviz_settings_attribute_in_dash_app
 
--- a/webviz_config/templates/webviz_template.py.jinja2
+++ b/webviz_config/templates/webviz_template.py.jinja2
@@ -20,6 +20,9 @@ from webviz_config.themes import installed_themes
 from webviz_config.common_cache import CACHE
 from webviz_config.webviz_store import WEBVIZ_STORAGE
 from webviz_config.webviz_assets import WEBVIZ_ASSETS
+from webviz_config.webviz_instance_info import WebvizRunMode
+from webviz_config.webviz_instance_info import WebvizInstanceInfo
+from webviz_config.webviz_factory_registry import WEBVIZ_FACTORY_REGISTRY
 from webviz_config.utils import deprecate_webviz_settings_attribute_in_dash_app
 
 # We do not want to show INFO regarding werkzeug routing as that is too verbose,
@@ -73,10 +76,21 @@ CACHE.init_app(server)
 theme.adjust_csp({"script-src": app.csp_hashes()}, append=True)
 Talisman(server, content_security_policy=theme.csp, feature_policy=theme.feature_policy, force_https=False, session_cookie_secure=False)
 
+storage_folder = Path(__file__).resolve().parent / "resources" / "webviz_storage"
+
 WEBVIZ_STORAGE.use_storage = {{portable}}
-WEBVIZ_STORAGE.storage_folder = Path(__file__).resolve().parent / "resources" / "webviz_storage"
+WEBVIZ_STORAGE.storage_folder = storage_folder
 
 WEBVIZ_ASSETS.portable = {{ portable }}
+
+run_mode = WebvizRunMode.PORTABLE if {{portable}} else WebvizRunMode.NON_PORTABLE
+app_instance_info = WebvizInstanceInfo(run_mode, storage_folder)
+{% if internal_factory_settings is defined %}
+factory_settings_dict = {{ internal_factory_settings }}
+{% else %}
+factory_settings_dict = None
+{% endif %}
+WEBVIZ_FACTORY_REGISTRY.initialize(app_instance_info, factory_settings_dict)
 
 use_oauth2 = False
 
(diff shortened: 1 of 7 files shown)


webviz_config/templates/webviz_template.py.jinja2
```diff
@@ -20,6 +20,9 @@ from webviz_config.themes import installed_themes
 from webviz_config.common_cache import CACHE
 from webviz_config.webviz_store import WEBVIZ_STORAGE
 from webviz_config.webviz_assets import WEBVIZ_ASSETS
+from webviz_config.webviz_instance_info import WebvizRunMode
+from webviz_config.webviz_instance_info import WebvizInstanceInfo
```
anders-kiaer, line 24:
```suggestion
from webviz_config.webviz_instance_info import WebvizRunMode, WebvizInstanceInfo
```

webviz_config/webviz_factory_registry.py
```diff
@@ -0,0 +1,84 @@
...
+            raise ValueError("Illegal access, factory registry is not initialized")
+
+        if not isinstance(factory_obj, factory_class):
+            raise TypeError("The type of the factory does not match factory_class")
+
+        self._factories[factory_class] = factory_obj
+
+    def get_factory(self, factory_class: Type[T]) -> Optional[T]:
+        if not self._is_initialized:
+            raise ValueError("Illegal access, factory registry is not initialized")
```
anders-kiaer, line 58: `RuntimeError` instead, or maybe a new custom error is worth it since "Factory registry is not initalized" can happen multiple places?

https://docs.python.org/3/tutorial/errors.html#user-defined-exceptions

webviz_config/webviz_factory_registry.py
```diff
@@ -0,0 +1,84 @@
...
+        self._app_instance_info = app_instance_info
+
+        if factory_settings_dict:
+            if not isinstance(factory_settings_dict, dict):
+                raise TypeError("factory_settings_dict must be of type dict")
+            self._factory_settings_dict = factory_settings_dict
+
+        self._is_initialized = True
+
+    def set_factory(self, factory_class: Type[T], factory_obj: T) -> None:
```
anders-kiaer, line 47: Have we considered getting the factory class from `.__class__`?

```python
class FactoryClass:
    pass

factory_obj = FactoryClass()

factory_obj.__class__ == FactoryClass  # == true
isinstance(factory_obj, FactoryClass)  # == true
```

anders-kiaer (changes requested):
Looks good. :clap: :+1: First review done. :rocket: 

Should we also add a section in `CONTRIBUTIONG.md` showing high level usage for plugin projects? Have we considered a base class for `FactoryClass`es or is not needed/useful? :slightly_smiling_face:

webviz_config/webviz_factory_registry.py
sigurdp: Maybe `RuntimeError` is best suited?
It doesn't seem worth it to introduce a custom exception since the only time this error should occur is if `WebvizFactoryRegistry` is used without being initialized. Normal usage of the registry is via `WEBVIZ_FACTORY_REGISTRY` and in that case our own jinja2 templates are resposnible for initialization.

webviz_config/webviz_factory_registry.py
sigurdp: As far as I understand that would not work here. The intention is that `factory_class` should serve as the lookup key in the registry and that the `factory object` should be an instance of `factory_class` or a subclass of `factory_class`. 
Wouldn't comparing against `.__class__` require an exact match?

sigurdp:
> Looks good. 👏 👍 First review done. 🚀
> 
> Should we also add a section in `CONTRIBUTIONG.md` showing high level usage for plugin projects? Have we considered a base class for `FactoryClass`es or is not needed/useful? 🙂

Did actually consider adding a `WebvizFactory` base class initially, but at the time there was no common interface requirements for the factories, so the idea was dismissed. *However*, I now think we may end up needing some common functionality for all factories. The first such requirement is probably an optional cleanup function that can be called on all factories to instruct them to release temporary cache memory once all plugins have been initialized. Something along the lines of `WebvizFactory.cleanupTempInitializationResources()`. 

So will add a `WebvizFactory` base class that all factories should derive from.

webviz_config/webviz_factory_registry.py
anders-kiaer: Ah, ok. Yes, `.__class__` would require an exact match.

From the (to be added) base class, would a given chain of subclasses have _one_ class in the chain to use as key?

Maybe `__bases__` or `__mro__` can be used? I'm probably of more review help after the `CONTRIBUTING.md` section with intended high-level usage in plugin projects. :smile: 

```python
class FactoryClass:
    pass

class SubFactoryClass(FactoryClass):
    pass

sub_factory_obj = SubFactoryClass()

print(sub_factory_obj.__class__.__mro__)  # (<class '__main__.SubFactoryClass'>, <class '__main__.FactoryClass'>, <class 'object'>)
print(sub_factory_obj.__class__.__bases__)  # (<class '__main__.FactoryClass'>,)
```

webviz_config/webviz_factory.py
```diff
@@ -0,0 +1,11 @@
+class WebvizFactory:
+    """Base class for all factories that want to register themselves in
+    WebvizFactoryRegistry.
+    """
```
anders-kiaer, line 4: Should this also be an ABC? I.e. it doesn't make sentence to create an instance of `WebvizFactory` directly?

webviz_config/webviz_factory_registry.py
```diff
@@ -0,0 +1,93 @@
...
+    def all_factory_settings(self) -> Dict[str, Any]:
+        if not self._is_initialized:
+            raise RuntimeError("Illegal access, factory registry is not initialized")
+
+        return self._factory_settings_dict
+
+    @property
+    def app_instance_info(self) -> WebvizInstanceInfo:
+        if not self._is_initialized or self._app_instance_info is None:
+            raise RuntimeError("Illegal access, factory registry is not initialized")
```
anders-kiaer, line 88:
```suggestion
            raise RuntimeError("Factory registry is not initialized")
```
This is perhaps enough error message feedback (stating the problem)?

webviz_config/templates/webviz_template.py.jinja2
```diff
@@ -73,11 +75,22 @@ CACHE.init_app(server)
...
 WEBVIZ_ASSETS.portable = {{ portable }}
 
+run_mode = WebvizRunMode.PORTABLE if {{portable}} else WebvizRunMode.NON_PORTABLE
+app_instance_info = WebvizInstanceInfo(run_mode, storage_folder)
+{% if internal_factory_settings is defined %}
+factory_settings_dict = {{ internal_factory_settings }}
+{% else %}
+factory_settings_dict = None
+{% endif %}
+WEBVIZ_FACTORY_REGISTRY.initialize(app_instance_info, factory_settings_dict)
```
anders-kiaer, line 92:
```suggestion
WEBVIZ_FACTORY_REGISTRY.initialize(app_instance_info, {{ internal_factory_settings if internal_factory_settings is defined else None }})
```
Better? https://stackoverflow.com/questions/14214942/jinja2-shorthand-conditional

anders-kiaer (commented):
LGTM! :rocket: Since we are not 100% sure how this API will end up, should we add comment somewhere that this is provisional/experimental, and will not necessarily see a deprecation phase and "may deviate from the usual version semantics"?

Even the Python std.lib does this for some modules: https://docs.python.org/3/library/importlib.metadata.html

webviz_config/webviz_factory.py
sigurdp: I did consider it, but I don't think it makes sense since there are currently no abstract methods. I believe that it is the presence of actual abstract methods that make the class abstract and therefore prohibits instantiation. In that case I don't think it helps making the class itself an ABC. Not sure though...

webviz_config/templates/webviz_template.py.jinja2
sigurdp: Definitively more compact

sigurdp:
> LGTM! 🚀 Since we are not 100% sure how this API will end up, should we add comment somewhere that this is provisional/experimental, and will not necessarily see a deprecation phase and "may deviate from the usual version semantics"?
> 
> Even the Python std.lib does this for some modules: https://docs.python.org/3/library/importlib.metadata.html

In your opinion, would it be sufficient to add such a disclaimer to the docstring of `WebvizFactoryRegistry`, or should we comment it elsewhere?

anders-kiaer:
> In your opinion, would it be sufficient to add such a disclaimer to the docstring of WebvizFactoryRegistry, or should we comment it elsewhere?

Yes, I think that should be sufficient (since it is not documented in "rendered docs"), maybe also in the `WebvizFactory` base class docstring?

anders-kiaer (approved):

PR #435, opened 2021-04-29
Support + test source_url/download_url combinations

### LINKED ISSUE #431: Download URL is needed if source URL is given in setup.py
If a plugin project has stated source url in `setup.py`, but not download url, it fails here:
https://github.com/equinor/webviz-config/blob/e42cc2df1165aae8d2d68c98b507e294bc36aaf0/webviz_config/_dockerize/_create_docker_setup.py#L63-L81

`AttributeError: 'NoneType' object has no attribute 'startswith'`

anders-kiaer:
Closes #431.


HansKallekleiv (approved):
LGTM 👍

PR #437, opened 2021-05-03
Modify connect-src CSP for DeckGL component

HansKallekleiv:

---

### Contributor checklist

   - [ ] Task 1
   - [ ] Task 2

### DIFF
--- a/webviz_config/_theme_class.py
+++ b/webviz_config/_theme_class.py
@@ -12,19 +12,19 @@ class WebvizConfigTheme:
 
         self._csp = {
             "default-src": "'none'",
-            "connect-src": "'self'",
+            "connect-src": ["'self'", "data:"],  # [3]
             "prefetch-src": "'self'",
             "style-src": ["'self'", "'unsafe-inline'"],  # [1]
             "script-src": [
                 "'self'",
                 "'unsafe-eval'",  # [2]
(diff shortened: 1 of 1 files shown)


anders-kiaer:
Interesting. I guess we might not want to open this up in the default CSP settings (since not all Webviz apps will be using `webviz-subsurface-components`). Even though a lot of web sites out there in general still don't use CSP at all, we should aim at keeping our CSP settings strict. Snippet from https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/connect-src:

> <scheme-source>
>     A scheme such as http: or https:. The colon is required. Unlike other values below, single quotes shouldn't be used. You can also specify data schemes (not recommended).
> 
>  * `data:` Allows `data:` URIs to be used as a content source. This is insecure; an attacker can also inject arbitrary data: URIs. Use this sparingly and definitely not for scripts.

Two possible ways forward:
1. Any other ways `webviz-subsurface-components` can go ahead with DeckGL without allowing `data :` in `connect-src`?
2. If not - we might want to investigate a way of plugin projects to set CSP settings based on their needs.

anders-kiaer:
For reference: https://github.com/niklasvh/html2canvas/issues/2521

PR #443, opened 2021-05-07
Improved configuration of logging granularity in webviz

### LINKED ISSUE #439: Improve logging granularity in webviz
Global loglevel for `webviz` can currently be specified on the command line when doing webviz build in this way:
`webviz build ../webviz-subsurface-testdata/webviz_examples/webviz-full-demo.yml --loglevel INFO`

This specifies the global, application-wide, loglevel. Using such a global loglevel of `INFO` or `DEBUG` will quickly flood the log with messages from all possible sub-modules. For verbose logging to be of any use we need finer grain control over the loglevels on a per-module basis.

We should implement a way of specifying loglevels on a per-module basis so that full verbosity can be enabled for modules of interest, while log messages from other modules remain muted. Proposed solution is to either modify the exisiting `--loglevel` argument or intoduce a new argument (`--logconfig` ?) that accepts a YAML file with logging configuration in a standardized format.

If we introduce a new argument, the priorities between the two mechanisms must be clarified. 

See:
https://docs.python.org/3/library/logging.config.html#logging.config.dictConfig
https://docs.python.org/3/library/logging.config.html#configuration-dictionary-schema

sigurdp:
Added `--logconfig` command line argument to allow finer grain control over logging output from webviz applications.

The logging configuration is specified using a YAML file that describes the logging setup as a dictionary according to the schema specified in: https://docs.python.org/3/library/logging.config.html#logging-config-dictschema. 
After the YAML has been parsed, the configuration dictionary will be applied using `logging.config.dictConfig()` (see https://docs.python.org/3/library/logging.config.html#logging.config.dictConfig)

---

### Contributor checklist

- [x] :tada: This PR closes #439


INTRODUCTION.md
```diff
@@ -132,6 +132,49 @@ If you are using Visual Studio Code, we recommend [Red Hat's YAML extension](htt
...
+
+In practice, setting the global logging level to `INFO` or `DEBUG` will likely flood the log with messages from various 
+sub-modules, possibly obscuring messages from the module(s) of interest. To allow finer grain control over logging, 
+Webviz accepts the `--logconfig` command line argument which allows detailed logging configuration through a YAML file:
+```bash
+webviz build ./examples/basic_example.yaml --logconfig my_log_config.yml
+```
+
+The YAML file is expected to contain a dictionary with logging configuration adhering to the [schema described 
+here.](https://docs.python.org/3/library/logging.config.html#logging-config-dictschema) This allows for full
```
anders-kiaer, line 151:
```suggestion
here](https://docs.python.org/3/library/logging.config.html#logging-config-dictschema). This allows for full 
```

INTRODUCTION.md
```diff
@@ -132,6 +132,49 @@ If you are using Visual Studio Code, we recommend [Red Hat's YAML extension](htt
...
+loggers:
+  werkzeug:
+    level: ERROR
+  some.other.module:
+    level: INFO
+```
+
+Please note that if both `--loglevel` and `--logconfig` are specified, the latter will take precedence. Internally, the 
+global log level set by `--loglevel` will be applied first. Then the configuration specified by `--logconfig` will
+be applied, possibly overwriting any overlapping settings.
```
anders-kiaer, line 176: Good explanation. :+1: What was perhaps a bit unclear from reading this is if also the `incremental` key is mandatory, and if not what happens if it is not included. Should we add one sentence regarding the difference incremental vs. non-incremental?

webviz_config/templates/copy_data_template.py.jinja2
```diff
@@ -41,6 +42,13 @@ app._deprecated_webviz_settings = {
     "theme" : webviz_settings.theme
 }
 
+{% if logging_config_dict is defined %}
+# Apply a logging config dict as specified via the --logconfig command line argument. 
+logging_config_dict = {{ logging_config_dict }}
+print("Configuring logging from dictionary:", logging_config_dict)
```
anders-kiaer, line 48: Do we want this to be printed, or is it remains after the development?

anders-kiaer (commented):
LGTM! Could only find a couple of minor things to comment on :male_detective:

webviz_config/templates/copy_data_template.py.jinja2
sigurdp: I think maybe we should remove it here in the 'copy' template, but I'm inclined to leave it in place in the 'main' template so that when running a portable build it will apear in the console that a logging dictionary is being applied.

INTRODUCTION.md
sigurdp: Will try and make it clear that the `incremental` key is part of the official Python schema so that those inclined can look it up there.

INTRODUCTION.md
```diff
@@ -132,6 +132,52 @@ If you are using Visual Studio Code, we recommend [Red Hat's YAML extension](htt
...
+This global logging level can be changed on the command line using the `--loglevel` argument like this:
+```bash
+webviz build ./examples/basic_example.yaml --loglevel INFO
+```
+
+In practice, setting the global logging level to `INFO` or `DEBUG` will likely flood the log with messages from various 
+sub-modules, possibly obscuring messages from the module(s) of interest. To allow finer grain control over logging, 
+Webviz accepts the `--logconfig` command line argument which allows detailed logging configuration through a YAML file:
+```bash
+webviz build ./examples/basic_example.yaml --logconfig my_log_config.yml
```
anders-kiaer, line 147: Should we keep a consistent file extension for YAML in the command example? `.yaml` or `.yml`?

anders-kiaer (approved):
LGTM! 🚀 One minor comment only.

INTRODUCTION.md
sigurdp: Absolutely. Since the `--logfile` parameter in the example refers to a imaginary file I'll just change the extension to `.yaml`

PR #446, opened 2021-05-10
Use latest (including pre-releases) of webviz-core-components in CI

anders-kiaer:
We are planning on starting with pre-releases on `webviz-*` packages to PyPI/npm. In order to be able to check consistency early in CI, we instruct the GitHub actions workflow to include prereleases of `webviz-core-components`.


.github/workflows/webviz-config.yml
```diff
@@ -46,11 +46,11 @@ jobs:
         run: |
           pip install --upgrade pip
           pip install . .[deployment]
+          pip install --upgrade webviz-core-components  # Testing against our latest prerelease
 
       - name: 📦 Install test dependencies
         run: |
           pip install .[tests]
-          pip install dash[testing]
```
anders-kiaer, line 53: Remains after #444.

rubenthoms (approved):
LGTM! :+1: Nice that we finally get rid of line 53.

PR #449, opened 2021-05-11
Adjusted Dash application logging to webviz's global logging mechanism.

sigurdp:
Enforce the user specified global log level for the Dash app logger instead of always setting it to WARNING.
Remove the log handler that Dash insits on adding by default. This handler should not be needed since a handler gets added when we call `logging.basicConfig()` and we currently end up with two handlers for the Dash app logger.
Hardwire Dash application name to 'webviz_app' when creating the application object (instead of using `__name__`)


anders-kiaer (approved):
LGTM :+1: :books:

PR #451, opened 2021-05-12
Disable pylint warning on `flask.g` object, and fix `mypy` issues after type hinted `flask`

### LINKED ISSUE #450: Pylint fails after new release of flask

anders-kiaer:
Based on feedback from https://github.com/pallets/flask/issues/4020.

Closes #450.


sigurdp (approved):
LGTM.
Should probably keep an eye out for Flask 2.0.1 and see if the disabled pylint warnings can be restored when it is released. See https://github.com/pallets/flask/pull/4042

PR #482, opened 2021-07-05
Docs: Reinstate free text for arg desc + minor restructure of deprecation folder

asnyv:
Existing look of arguments (from webviz-subsurface):
image

New look (the argument definition is now back to basically a free text field as before the deprecation feature was added):
image

Existing deprecation folder structure. Includes deprecated arguments, not only plugins, so a bit misleading + there are two levels of _Deprecated plugins_:
image

New structure (switched from _plugin project_ to _package_ to keep in line with the rest of the menu:
image

I removed the code related to extracting documentation of individual arguments with regex that was added as part of #419 (mainly the part [here](https://github.com/equinor/webviz-config/blob/e42cc2df1165aae8d2d68c98b507e294bc36aaf0/webviz_config/_docs/_build_docs.py#L88) related to `documented_args`, as this code is based on a strict structure of a single list of arguments with descriptions, which a lot of plugins are not adhering too (if such a feature is to be included, we have to make it a requirement and have a way to add additional information that doesn't naturally belong to a single argument)

Look of deprecated arguments after the update (also added `Deprecated` as a possible comment in the yaml description):
image

---

### Contributor checklist

   - [X] Reinstate plugin docstring section for arguments as a free text field
   - [X] Restructure folder in docs for deprecations


asnyv:
I got some more "ideas", so consider this a draft as of now

asnyv:
Separated deprecated plugins and deprecated arguments into two sections in the sidebar, as both were placed under _Deprecated plugins_ and the listed plugins in the menu could then confuse users to believe that whole plugins were deprecated, when only an argument is.

Also separated the code to generate the single plugin docs into a new jinja-file (rather than having to maintain one for deprecations and one for the regular doc.
image

CONTRIBUTING.md
```diff
@@ -642,14 +642,16 @@ class MyPluginExample1(WebvizPluginABC):
...
-    def __init__(self, arg1: str, arg2: int, arg3: Optional[int] = None):
+    @deprecated_plugin_arguments(check_deprecation_example2)
+    def __init__(self, arg1: int, arg2: int, arg3: int):
         ...
 
-def check_deprecation(arg1: int, arg3: int) -> Optional[Tuple[str, str]]:
-    if arg3 == arg1:
-        return ("This message is shown to the end user in the app.", "This message is shown in the documentation of the plugin.")
-    return None
+
```
anders-kiaer, line 656: Necessary empty line?

anders-kiaer (approved):
Nice, LGTM! :rocket:

CONTRIBUTING.md
asnyv: Probably just a consequence of me cutting the code block and pasting it further up, leaving the empty line. I can remove it

PR #489, opened 2021-08-25
Support Radix non-root and new blob driver

### LINKED ISSUE #477: `FlexVolume` deprecated by Microsoft and Radix
Replaced by Azure Blob Storage CSI driver. In our context, it approx. only means replacing some argument names in the auto-generated Radix configuration.

See https://www.radix.equinor.com/docs/reference-radix-config/#volumemounts

### LINKED ISSUE #486: User numeric format of non-root Docker user
The `webviz` Docker app currently runs as user `appuser` (i.e. non-root). To be Radix compatible we however need to use user id on numerical form (as Radix don't read `appuser` correctly as a non-root user).

anders-kiaer:
### Contributor checklist

- [X] :tada: This PR closes #477 and closes #486.


asnyv (approved):
LGTM

PR #498, opened 2021-09-09
When comparing timezones, explicitly use UTC.

### LINKED ISSUE #497: "TypeError: can't compare offset-naive and offset-aware datetimes" when attempting to run with oauth enabled
python version: 3.8
webviz-config version: 0.3.2

When attempting to run a webviz server with oauth enabled, the server encounters an internal error. The error message originates from checking whether the oauth token has expired:
```
  File "/.../venv/lib/python3.8/site-packages/webviz_config/_oauth2.py", line 133, in check_and_set_token_expiry
    if current_date > expiration_date:
TypeError: can't compare offset-naive and offset-aware datetimes
```

The origin of the error appears to be the storage of the expiration date as a session object. When the datetime gets stored as a session object, it gets serialized into a timezone-aware object (see https://github.com/pallets/flask/blob/0.10.1/flask/sessions.py#L73 (old version of flask, but implementation detail is the same on more recent versions) and https://werkzeug.palletsprojects.com/en/2.0.x/http/).

We can fix the symptom by making sure that we explicitly use datetime objects with a utc date.

mhwaage:
If not setting a timezone, the timezone will be timezone naive. However, this will get serialized to a timezone-aware
timezone when being stored as a session object. This causes later comparisons to fail, since
we end up compariong timezone aware to timezone naive objects. Instead, always explicitly create timezone-aware datetimes

### Contributor checklist

- [x] :tada: This PR closes #497 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [UNRELEASED] - YYYY-MM-DD
 
+### Fixed
+- [#498](https://github.com/equinor/webviz-config/pull/498) - Explicitly use UTC when comparing time zones in the `OAuth` implementation.
+
 ## [0.3.4] - 2021-08-30
 
 ### Fixed
--- a/webviz_config/_oauth2.py
+++ b/webviz_config/_oauth2.py
@@ -77,7 +77,7 @@ class Oauth2:
             expiration_date = datetime.datetime.now(
                 datetime.timezone.utc
             ) + datetime.timedelta(seconds=expires_in - 60)
-            print("Access token expiration date:", expiration_date)
+            print("Access token expiration date (UTC):", expiration_date)
 
             # Set expiration date in the session
             flask.session["expiration_date"] = expiration_date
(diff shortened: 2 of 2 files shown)


mhwaage:
wrt testing, as far as I can tell there are no tests for the oauth implementation currently. Adding it would require some amount of mocking, I am not sure if that would be considered in scope?

anders-kiaer (approved):
LGTM, thanks @mhwaage 🚀  If @thezultimate also gives 👍 we can merge this as is.

Perhaps this is worth a changelog entry? E.g. add 
```
### Fixed
- [#498](https://github.com/equinor/webviz-config/pull/498) - Explicitly use UTC when comparing time zones in the `OAuth` implementation.
```
under the `[UNRELEASED]` section in [`./CHANGELOG.md`](https://github.com/equinor/webviz-config/blob/master/CHANGELOG.md).

thezultimate:
Thanks @mhwaage for the PR of the fix. I managed to reproduce the issue. It worked when I used an old version of Flask (version 1.1.2), but failed when I used the latest version of Flask (version 2.0.1). I will continue testing in the morning to see if the automatic token refresh works properly with the new code. Please be patient :)

thezultimate (approved):

thezultimate:
Works nicely. Approved :)

PR #502, opened 2021-09-12
Bump to `dash>=2.0`

HansKallekleiv:
   - [ ] Bump Dash dependency to 2.0
   - [ ] Simplify import statement as suggested in [Dash 2.0 migration](https://dash.plotly.com/dash-2-0-migration)


anders-kiaer (commented):
👍 LGTM. Should we include `orjson` as a new dependency in `webviz-config`?

https://dash.plotly.com/performance
> New with Dash 2.0, you can use `orjson` to speed up serialization to JSON and in turn improve your callback performance. For an app that runs complex callbacks involving a large amount of data points, you could see a performance increase of up to 750 ms. Learn more about `orjson` here or install it with `pip install orjson`.
>
> `orjson` is completely optional: If it exists in the environment, then Dash will use it; if not, Dash will use the default `json` library.

anders-kiaer (approved):
🚀

PR #527, opened 2021-09-30
Remove `UserWarning` from `dash-bootstrap-components`

### LINKED ISSUE #525: Convert `UserWarning` emitted from `dash==2.0` to `DeprecationWarning`
...since that is what it should have been in the first place. 

There are some libraries out there that haven't released yet a Dash2.0 compliant package, emitting things like:
```
.../dash_bootstrap_components/_table.py:5: UserWarning: 
The dash_html_components package is deprecated. Please replace
`import dash_html_components as html` with `from dash import html`
  import dash_html_components as html
```
This confuses non-technical users.

Try capture these and reemit them as `DeprecationWarning` (which is not shown by default)?

anders-kiaer:
Closes #525.

### DIFF
--- a/webviz_config/plugins/__init__.py
+++ b/webviz_config/plugins/__init__.py
@@ -2,7 +2,7 @@
 the utility itself.
 """
 
-from typing import Optional
+import warnings
 
 try:
     # Python 3.8+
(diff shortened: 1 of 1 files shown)


webviz_config/plugins/__init__.py
```diff
@@ -2,7 +2,7 @@
 the utility itself.
 """
 
-from typing import Optional
```
anders-kiaer, line 5: I'm a bit surprised `pylint` haven't complained about this... looks unused.

asnyv (approved):
🆗

PR #543, opened 2021-10-25
Improved image spacing

rubenthoms:
Set margin around image and image caption (`_markdown_image_caption`) to `16px` each.


webviz_config/static/assets/webviz_config.css
```diff
@@ -22,12 +22,15 @@
...
 
 }
 
 ._markdown_image_caption {
 
     display: block;
     text-align: center;
-
+    margin-top: 16px;
+    margin-bottom: 16px;
```
anders-kiaer, line 35: Do you have a screenshot of how this looks readily available? 🙂 Should we let `margin-top < margin-bottom` for the image caption, i.e. that the image caption text is closer to the image it belongs to compared with next element below, or is this already the case?

anders-kiaer (commented):
LGTM! One question to consider and then this is ready to go IMO. 👍

anders-kiaer (approved):
👍

PR #555, opened 2021-11-30
Wait until plugins are initiated before setting csp

### LINKED ISSUE #554: Clientside callbacks prevented by CSP
**Describe the bug**
Clientside callbacks are created by introducing inline scripts. Those inline scripts have associated hashes that we are whitelisting. Unfortunately, the whitelisting of hashes happens _before_ plugins are initiated, meaning that none of the callbacks are whitelisted.

**How to reproduce**
Make any plugin with a clientside callback and build it.

mhwaage:
This resolves #554


anders-kiaer (approved):
Thanks @mhwaage, LGTM! 👍🚀

PR #562, opened 2021-12-06
Added global `WebvizInstanceInfo`

### LINKED ISSUE #249: Expose instance of `WebvizRunMode` during build and runtime

sigurdp:
Added global `WebvizInstanceInfo` accessible through `WEBVIZ_INSTANCE_INFO.` 

An instance of `WebvizInstanceInfo` will be initialized/populated early on during a Webviz application's execution and can be accessed globally through `WEBVIZ_INSTANCE_INFO.` 

It has the following properties:
```
WebvizInstanceInfo.dash_app
WebvizInstanceInfo.run_mode
WebvizInstanceInfo.theme
WebvizInstanceInfo.storage_folder
```
Also marks `WebvizFactoryRegistry.app_instance_info` as deprecated

---

### Contributor checklist

- [X] :tada: This PR partially closes #249.
- [X] :robot: I have added tests, or extended existing tests, to cover any new features or bugs fixed in this PR.

### DIFF
--- a/webviz_config/webviz_instance_info.py
+++ b/webviz_config/webviz_instance_info.py
@@ -73,21 +73,21 @@ class WebvizInstanceInfo:
 
     @property
     def dash_app(self) -> Dash:
-        if not self._is_initialized:
+        if not self._is_initialized or self._dash_app is None:
             raise RuntimeError("WebvizInstanceInfo is not yet initialized")
 
         return self._dash_app
 
     @property
     def run_mode(self) -> WebvizRunMode:
-        if not self._is_initialized:
+        if not self._is_initialized or self._run_mode is None:
             raise RuntimeError("WebvizInstanceInfo is not yet initialized")
 
         return self._run_mode
 
     @property
     def theme(self) -> WebvizConfigTheme:
-        if not self._is_initialized:
+        if not self._is_initialized or self._theme is None:
             raise RuntimeError("WebvizInstanceInfo is not yet initialized")
 
         # The theme class is mutable, so return a copy
@@ -95,7 +95,7 @@ class WebvizInstanceInfo:
 
     @property
     def storage_folder(self) -> Path:
-        if not self._is_initialized:
+        if not self._is_initialized or self._storage_folder is None:
             raise RuntimeError("WebvizInstanceInfo is not yet initialized")
 
         return self._storage_folder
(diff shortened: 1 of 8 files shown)


sigurdp:
Should we also try and flag the magic `app` plugin parameter as deprecated now to try and nudge users into using the new structure, or should we hold back a bit?

anders-kiaer:
> Should we also try and flag the magic app plugin parameter as deprecated now to try and nudge users into using the new structure, or should we hold back a bit?

We could perhaps deprecate it now (I guess we are pretty sure we don't want to go back to the "special `__init__` arguments approach).

webviz_config/webviz_instance_info.py
```diff
@@ -9,16 +15,90 @@ class WebvizRunMode(Enum):
 
 
 class WebvizInstanceInfo:
-    """Class containing global info regarding the running webviz app instance"""
+    """Containins global information regarding the running webviz app instance, exposed
```
anders-kiaer, line 18:
```suggestion
    """Contains global information regarding the running webviz app instance, exposed
```

webviz_config/webviz_instance_info.py
```diff
@@ -9,16 +15,90 @@ class WebvizRunMode(Enum):
...
+
+        if not isinstance(storage_folder, Path):
+            raise TypeError("storage_folder must be of type Path")
         self._storage_folder = storage_folder
 
+        self._is_initialized = True
+
+    @property
+    def dash_app(self) -> Dash:
+        if not self._is_initialized or self._dash_app is None:
```
anders-kiaer, line 76: Are the extra `is None` checks in the different `@property` necessary?

anders-kiaer (commented):
Very good @sigurdp 👍 A couple of comments below + add a changelog entry? Then ready for merge IMO.

webviz_config/webviz_instance_info.py
sigurdp: You're right, they should not be needed since they are guarded in `__init__`.
Good find - will fix

sigurdp: Tried to remove them, but then that causes issues with mypy/type checking since it can no longer verify that the return types will always match. For now I will just reintroduce the `is None` checks since they solve the typing issue and they don't make that much noise.

anders-kiaer (approved):
👍

PR #566, opened 2021-12-23
Remove `dash-bootstrap-components` `UserWarning` filter

anders-kiaer:
...since `dash-bootstrap-components` now have a release using the new `dash` import order (which ensures that the end user doesn't get `UserWarning`'s the user can't fix).


anders-kiaer:
This PR reverts #527.

Issue #574, opened 2022-01-28
Radix issues


HansKallekleiv:
- Blob files are kept from old deployments. Most severe for caching as a previously, possible wrong, cached result could be used.
- Use of e.g. `uuid4` for creating component ids breaks down when running multiple replicas (pod instances)
  --> We must create ids in a different way, e.g. by using the plugin class name and a counter for each instance of the plugin.

PR #575, opened 2022-01-30
Generate plugin ids from class instance instead of uuid4

HansKallekleiv:
Generate plugin ids on the form:
<some_string>-<plugin_class_name>-<instance_counter>
instead of:
<some_string>-<uuid4()>
Relates to #574

### DIFF
--- a/webviz_config/_plugin_abc.py
+++ b/webviz_config/_plugin_abc.py
@@ -5,7 +5,6 @@ import zipfile
 import warnings
 import sys
 import urllib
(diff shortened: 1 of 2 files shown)


anders-kiaer:
Similar to PR #260 - with the same problem perhaps in upstream Dash that we should check/discuss again if still present and/or blocking?

HansKallekleiv:
Fixed in https://github.com/equinor/webviz-config/pull/599

Issue #578, opened 2022-02-14
Refactoring: New layout, design and data storage features


rubenthoms:
Based on the discussion between @anders-kiaer, @HansKallekleiv, @jorgenherje and me, here is a short summary of the results:

**Short term changes**
- [X] Icon used for collapsing expanded views should be changed to 
![image](https://user-images.githubusercontent.com/69145689/153868452-8b2d1260-4936-4474-8bfe-fb1723843bee.png)
- [X] Switch from `bootstrap's` modal view to `MaterialUI's` draggable dialog (https://mui.com/components/dialogs/#draggable-dialog) 
- [x] Implement a wrapper around each view in a plugin in order to provide an expand mechanism

**Medium term changes**
- [x] Make settings related to a particular view available via a gear icon button in the view's wrapper options
![image](https://user-images.githubusercontent.com/69145689/153870126-8649ad36-e3d7-456b-aaf1-ae433749cdeb.png)
- [x] Consider replacing expand/collapse wrapper with [`Accordion`](https://mui.com/components/accordion/), [eds also have one](https://eds-storybook-react.azurewebsites.net/?path=/docs/surfaces-accordion--simple)
- [x] Move plugin settings to a drawer that can be pinned (like the main menu)
- [x] With a settings menu that has ownership to view layout, we can limit the current loading indicator to only relevant view (using e.g. functionality from https://dash.plotly.com/loading-states), and remove the global app loading indicator (https://github.com/equinor/webviz-config-equinor/blob/15f159132920314c48223421ad66a8bb80eca85a/webviz_config_equinor/assets/equinor_theme.css#L103-L131).

**Long term changes**
- [ ] Use URL to share the exact same view with all set parameters (ask Dash/Plotly to implement a URL sharing feature)
- [x] Make it possible to implement a plugin group, i.e. one plugin with the same data but with different views that share the data - the views can either be accessible as subpages in the menu or as tabs in the plugin
    - Should tabs only be used when settings are shared between different views and subpages if only data but no settings are shared?
- [ ] Store view settings in browser's client-side storage (e.g. menu's pinned state or expand/collapse state of settings' groups) https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Client-side_web_APIs/Client-side_storage

jorgenherje:
Good summary! Only one question: should the last bulletpoint `Move plugin settings to a drawer that can be pinned (like the main menu)` be a short term or medium term task instead?

[...]

PR #599, opened 2022-06-16
Introducing Webviz Layout Framework (WLF)

rubenthoms:
Based on https://github.com/equinor/webviz-config/issues/578, this PR introduces all Python classes and changes to the application template required by the new Webviz Layout Framework (WLF). That is (not exclusively):
- Fundamental new functionality in `_plugin_abc.py` while not introducing breaking changes
- New plugin subclasses:
  - `_layout_base_abc.py`: A base class for all new WLF classes
  - `_layout_unique_id.py`: A class for handling unique IDs for all new WLF instances
  - `_settings_group_abc.py`: A base class for implementing settings groups
  - `_views.py`: Base classes for view elements, layout elements and views
- Changes to `webviz_template.py.jinja2`: Implementing new structure with React components from https://github.com/equinor/webviz-core-components/pull/219 
- Added example plugins with WLF in `example_content_wrapper_plugin.py`

ℹ️  This PR depends on https://github.com/equinor/webviz-core-components/pull/219


HansKallekleiv (approved):
Very nice work! Looks great!

Couple of things that we should add as issues to solve later:
- References to `uuid` should be removed as plugin ids are not generated based on this anymore.
- Option to keep multiple settings groups open at the same time as discussed.

PR #606, opened 2022-08-05
Replace plugin wrapper id string with plugin unique id string in PluginABC

jorgenherje:
Remove unnecessary plugin wrapper id string from PluginABC - rather utilize plugin unique id string


jorgenherje:
Closed - handled in https://github.com/equinor/webviz-config/pull/607

PR #610, opened 2022-08-10
Bump Python minor version to `3.8`

anders-kiaer:


webviz_config/_config_parser.py
```diff
@@ -103,22 +103,18 @@ def _call_signature(
...
-                        f"of type `{type(kwargs[arg]).__name__}`. "
-                        f"Expected type "
-                        f"`{argspec.annotations[arg].__name__}`. "
-                        "Run the command `webviz docs` if you want "
-                        "to see documentation of the plugin."
-                        f"{terminal_colors.END}"
-                    )
-            # Typechecking typing classes does not work in python 3.7
-            except TypeError:
-                pass
```
anders-kiaer, line 121: @HansKallekleiv Safe to remove this now when we are on Python 3.8+? I see it was introduced in #153.

HansKallekleiv (approved):

PR #630, opened 2022-09-19
Improved `callback_typecheck` exception message

rubenthoms:
Error message does now include parameter, function and file names.


jorgenherje (changes requested):
Seems to be a couple of unused imports. Otherwise looks good to me!

jorgenherje (approved):

HansKallekleiv (approved):

PR #632, opened 2022-09-22
Improved typechecking, especially for `Union`

rubenthoms:
Introduced a new custom `_isinstance` function in `_callback_typecheck.py` in order to check if a variable already complies with the expected type. This does especially improve the type checking of `Union` types.

### DIFF
--- a/webviz_config/utils/_callback_typecheck.py
+++ b/webviz_config/utils/_callback_typecheck.py
@@ -1,5 +1,5 @@
 # pylint: disable=line-too-long
-from typing import Any, Callable, get_origin, _TypedDictMeta, TypeVar, Union  # type: ignore[attr-defined]
+from typing import Any, Callable, get_args, get_origin, _TypedDictMeta, TypeVar, Union  # type: ignore[attr-defined]
 import inspect
 
 T = TypeVar("T")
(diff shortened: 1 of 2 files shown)


webviz_config/utils/_callback_typecheck.py
```diff
@@ -9,12 +9,44 @@ class ConversionError(Exception):
...
+        result = True
+        for annotation_arg in arg:
+            result &= _isinstance(annotation_arg, get_args(annotation)[0])
+        return result
+
+    if get_origin(annotation) is dict and isinstance(arg, dict):
+        result = True
+        for key, value in arg.items():
+            result &= _isinstance(key, get_args(annotation)[0])
+            result &= _isinstance(value, get_args(annotation)[1])
```
anders-kiaer, line 38: The `&=` syntax was new to me. What does it do here? And is the `result` assigned on line 37 used somewhere?

anders-kiaer (approved):
LGTM. A small question from my side. 🙂

webviz_config/utils/_callback_typecheck.py
rubenthoms: `&=` is a `bitwise and` - it compares two arrays of bits and returns a new one holding `1` where both array values were `1` and `0` where either one or both of them were `0`.

```python
a = 00000101
b = 00000011

a & b = 00000001

a = 1 # True
b = 0 # False

a & b = 0

a = 1 # True
b = 1 # True

a & b = 1
```

One could also write
```python
result = result and _isinstance(key, get_args(annotation)[0])
```

PR #636, opened 2022-10-03
Fixed issue when `options` not given in config file

rubenthoms:
Webviz would not build when no `options` were given in config file. This was due to `homepage` property missing in fallback `options` dictionary.


asnyv (approved):
LGTM 👍

PR #639, opened 2022-10-04
Fixed issue with `Any`, empty `Dict`/`List` and `None` in `callback_typecheck`

rubenthoms:
Also adjusted tests in order to test for these cases.

### DIFF
--- a/tests/test_callback_typecheck.py
+++ b/tests/test_callback_typecheck.py
@@ -1,4 +1,4 @@
-from typing import Dict, List, Optional, TypedDict, Union
+from typing import Any, Dict, List, Optional, TypedDict, Union
 from enum import Enum
 from webviz_config.utils import callback_typecheck, ConversionError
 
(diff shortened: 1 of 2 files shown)


anders-kiaer (approved):

lindjoha:
This does not solve the issue: https://github.com/equinor/webviz-subsurface/pull/1129. It still does not work to use `Optional[Dict[str, Any]]` as type hinting in the _update_graph() callback.

lindjoha:
> This does not solve the issue: [equinor/webviz-subsurface#1129](https://github.com/equinor/webviz-subsurface/pull/1129). It still does not work to use `Optional[Dict[str, Any]]` as type hinting in the _update_graph() callback.

After making sure that the correct version of webviz-config was used locally it turned out that it did actually solve the problem. https://github.com/equinor/webviz-subsurface/pull/1129

lindjoha (approved):
This solved the problem with `Optional[Dict[str, Any]]` I had in the `WellAnalysis` plugin.

PR #648, opened 2022-11-21
Allow `blob:` in `script-src` CSP

anders-kiaer:
Allow `blob:` in `script-src` CSP in order to enable web worker usage in Dash components.

Computational heavy Dash components might want to use web workers. Since it is not easy in Dash to host related worker scripts, these worker scripts are typically created on demand and given as `blob:` URLs. We do not allow `data:` script URLs since they do not have the same security feature as `blob:` URLs which can only be by created by the browser itself and have scope limited to the windows it was created in.


HansKallekleiv (approved):
LGTM :+1:

PR #655, opened 2022-11-24
Regression fix: Show filters in TablePlotter when using `lock`

HansKallekleiv:
Regression fix of https://github.com/equinor/webviz-config/pull/635

Tableplotter plugins using the `lock` arguments should still display any specified filters.


anders-kiaer (approved):
LGTM 👍

PR #663, opened 2023-01-09
Do not silently refresh tokens

### LINKED ISSUE #662: In scenarios with multiple users, oauth refresh tokens will return tokens belonging to the first user that logged in
In scenarios where the webviz solution has been deployed to a web server and serves multiple end users, it incorrectly handles assignment of refresh tokens.
Specifically, when attempting to refresh access tokens, it will use the refresh token belonging to the first account that signed in for that deployment. Concretely, consider the following scenario

* Application is deployed
* User A signs in
* User B signs in
* (1 hour passes)
* User A is still interacting with the application, gets issued a new access token from refresh token
* User B is still interacting with the application, gets issued a new access token from refresh token

In this scenario, User B will now be issued with an access token belonging to User A. Since these tokens can be used to control what users have access to see, it represents a security risk.

For a simple immediate elimination of the security risk, the current refresh mechanism should be abandoned and replaced with a forced interactive re-login. Unfortunately, this will cause issues with the user friendliness of the applications, as it means users will lose state every hour.

mhwaage:
this is a hotfix for a security issue where silent token refresh asigns the wrong tokens to end users. A significant drawback of this hotfix is that it now forces an interactive token-retrieval every time tokens expire, which will cause the user to lose work in progress (unless state is set to persist)

---

### Contributor checklist

- [x ] :tada: This PR closes #662.
   - [ ] Task 1
   - [ ] Task 2

### DIFF
--- a/webviz_config/_oauth2.py
+++ b/webviz_config/_oauth2.py
@@ -151,6 +151,10 @@ class Oauth2:
                 # Access token has expired
                 print("Access token has expired, trying to refresh.")
                 try:
+                    raise Exception(
+                        "Disable refreshing, it is broken and a security risk"
+                    )
+                    # pylint: disable=W0101
                     access_token, expiration_date = self.refresh_token_if_possible()
                     flask.session["access_token"] = access_token
                     flask.session["expiration_date"] = expiration_date


anders-kiaer (approved):
Thanks @mhwaage 🎉❤️ LGTM.

PR #665, opened 2023-01-20
fix: use a pure path rather than real path when defining config path …

### LINKED ISSUE #548: Portable application built on Windows fails in Radix due to WindowsPath instantiation
**Describe the bug**
When a Webviz portable application built on Windows platform is deployed to the Radix platform, the "main" component fails with `Error: cannot instantiate 'WindowsPath' on your system` message.

**How to reproduce**
1. Create and test a Webviz custom plugin project
2. On a Windows PC, build a portable application
3. Deploy the portable application to Radix platform, e.g. "Playground" environment
4. See error in the application's Radix console

**Expected behavior**
No error is given when application is built on Windows platform, or clear instruction NOT to use Windows for building apps to be deployed to Radix

**Screenshots**
image

mhwaage:
…in templates

real paths will throw if they are instantiated on the wrong system, which causes protable builds made in windows to fail when deploying in linux containers, e.g. on radix

---

### Contributor checklist

- [x] :tada: This PR closes #548.
   - [ ] Task 1
   - [ ] Task 2

### DIFF
--- a/webviz_config/_config_parser.py
+++ b/webviz_config/_config_parser.py
@@ -195,7 +195,9 @@ def _call_signature(
         special_args += "webviz_settings=webviz_settings, "
 
     return (
-        f"{plugin_name}({special_args}**{kwargs})",
+        f"{plugin_name}({special_args}**{kwargs})".replace(
+            "WindowsPath", "Path"
+        ).replace("PosixPath", "Path"),
         (
             f"plugin_layout(contact_person={contact_person}"
             f", plugin_deprecation_warnings={plugin_deprecation_warnings}"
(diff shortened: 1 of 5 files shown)


webviz_config/_config_parser.py
```diff
@@ -194,8 +194,8 @@ def _call_signature(
     if "webviz_settings" in argspec.args:
         special_args += "webviz_settings=webviz_settings, "
 
-    return (
-        f"{plugin_name}({special_args}**{kwargs})",
+    return  (
```
mhwaage, line 197: I am not very happy with this, but unsure of a good solution here - not sure if it is worth e.g. looping over the kwargs to more explicitly special-case Paths, or if this is good enough - especially given the long term plans of webviz

anders-kiaer: We will probably move `webviz-config` over to maintenance mode going forward (since the FMU usage pattern, funding this repo, is changing), and instead focus on collaborating with common React (and corresponding Dash) components.

I.e. I think your fix is reasonable given that context. 👍

anders-kiaer (approved):
LGTM 👍

PR #671, opened 2023-08-02
Fix JSONDecodeError to get CI running

tnatt:
Remove end slash from `PYPI_URL_ROOT` to avoid JSONDecodeError in `requests` due to double slashes

`PYPI_URL_ROOT = "https://pypi.org"`
`pypi_data = requests.get(f"{PYPI_URL_ROOT}/pypi/{dist_name}/json").json()`


asnyv (approved):
LGTM 😊

PR #697, opened 2026-05-11
Using `azure-monitor-opentelemetry-exporter`

anders-kiaer:
The package `azure-monitor-opentelemetry` package has a quite big dependency tree. In order to easier support deployment to Komodo, switching to `azure-monitor-opentelemetry-exporter`


sigurdp (approved):
LGTM 👍

